PR #12, opened 2024-03-16
Add u32_mul

weikengchen:
This PR presents an implementation of the u32_mul, which can be used for ASM_MUL, for 32-bit integers.

It implements:
- u8_to_bits, convert a u8 into bits
- u32_to_bits, convert a u32 (represented as four u8 in the stack) into bits
- u32_double, double a u32
- u32_mul_drop, multiplication of two u32

And their integration tests.

The instruction deals with overflow and resembles RISC-V behavior. It only maintains the keeps the lower 4 bytes of the 32 bit integers. It is possible to similarly implement 64-bit integer operations, but it depends on the ISA.

  u32_mul_drop takes 6118 bytes
  u32_add takes 85 bytes

roughly 72x, which is expected because u32_mul_drop calls u32_add already 64 times for double-and-add

### DIFF
--- a/src/scripts/opcodes/mod.rs
+++ b/src/scripts/opcodes/mod.rs
@@ -9,6 +9,7 @@ pub mod u256_std;
 pub mod u32_add;
 pub mod u32_and;
 pub mod u32_cmp;
+pub mod u32_mul;
 pub mod u32_or;
 pub mod u32_rrot;
 pub mod u32_state;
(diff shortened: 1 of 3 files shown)


weikengchen:
A new implementation is added, which reduces the number of bytes to 3705.

cahngdoudoulove:
great

PR #20, opened 2024-03-18
chore: fmt&clippy

Ethan-000:
mostly formatting and clippy fixes :) 

~except line 39 and 96 of bitvm.rs. can't build without adding `as u32` because pushable is not implemented for `u8`~ (resolved in main in new commit)


lucidLuckylee:
Sorry already outdated. I also want to keep some of the formatting (e.g. in merkle.rs).
Further, it is easier for us if we just run formatters ourselves since we don't have to review your code then and make sure there were only style changes. 
Thus, I generally won't merge automated fmt and clippy changes into the codebase.

Ethan-000:
hmm can you guys add a formatting ci or something. i started with this because i have fmt on save. if i dont do this and write other code there will be random formatting everywhere

lucidLuckylee:
We will eventually but it doesn't have priority. You can stage specific hunks in git instead of entire files if you want to keep your format on save enabled to avoid commiting format changes.

Ethan-000:
ahh ok

PR #28, opened 2024-03-26
add uint double/1add and u1, u2, ..., u30 to_bits

weikengchen:
This PR adds the double function for uint, as well as a generic implementation of to_bits for 30-bit limbs (or fewer). The to_bits implementation intentionally tries to avoid pushing the 2^k in the script in order to save weight units.


weikengchen:
I also reorganized the files into a subfolder `ubigint`, still as a submodule, to make it easier to organize the codes---I am going to add karatsuba soon.

PR #36, opened 2024-03-29
Alternative winternitz implementation, no digits on input.

tomkosm:
This is an alternative implementation, the idea is that we might save stack input space by calculating the digits, instead of adding them on the stack as input.


RobinLinus:
Great work! Thank you very much

PR #41, opened 2024-04-03
optimize fq2 and fq6 square

Hakkush-07:
This PR further optimizes Fq2::square from 749135 bytes to 500264 bytes and also optimizes Fq6::square from 4529796 bytes to 3022200 bytes.


RobinLinus:
solid improvement!! 🔥🔥🔥

weikengchen:
Wow

PR #43, opened 2024-04-03
add div2 and optimize fq6::square

Hakkush-07:
This PR adds functions for dividing by 2 in Fq and Fq2, and improves Fq2::square by using it from 3022200 bytes to 2782732 bytes.


RobinLinus:
Nice improvement! Thank you very much!!

PR #49, opened 2024-04-08
improve u30_div3_carry and u30_shr1_carry

Hakkush-07:
This PR improves u30_div3_carry and u30_shr1_carry, also gets rid of some redundancies. This results in minor improvements in some div2, div3, mul, inv, square functions of Fq, Fq2, Fq6, Fq12.


weikengchen:
Good catch on the "1 OP_SUB"

PR #51, opened 2024-04-12
add sha256:a rust native version. 

yimingWOW:
**add a rust native version sha256. i'll translate it into bitvm operation next week.** 

 The next job seemed to take me back to college when i studying the compilation course, that's really boring. 
 
And i noticed that u have a js and a rust sdk. i have a idea but i don't know if it's feasible. 
Maybe u can  t develop a translation scaffolding, which could translate assembly language into btc script. Then we don't need to provide so many sdk with different program languages. u know, It's easy to convert high-level languages to assembly.


aaronkzhou:
nice one

RobinLinus:
Thank you for your submission! Please implement the Script version before opening a PR.

PR #52, opened 2024-04-12
Optimize u32 rrot

wz14:
Hi, this PR adds all u32 rrots operations from 0 to 31, and corresponding test case.
Length of new bitcoin script for the `u32_rrot12` (291 byte) is a little bit better than the original one (292 byte), so I remove the old one.
Other rrot operations are reserved.

I hope this pr will help some guys to implement sha256 in bitvm.


RobinLinus:
Thank you very much!! great job!

PR #60, opened 2024-04-19
optimize u32 and xor or not

Hakkush-07:
This PR optimizes u32 AND XOR OR NOT functions.

```
u32 and cost 170 instead of 174
u32 or cost 186 instead of 326
u32 xor cost 202 instead of 206
u32 not cost 24 instead of 230
```

It helps reduce the size of sha256 script.


RobinLinus:
thanks!! nice improvements!

PR #65, opened 2024-04-26
Optimized Sha256 version working with u4 

jonasmartin:
Hi! 
I've been working on a new version of the sha256 on-chain script that works using u4 as basic unit size.

In the process of doing it I created some interesting optimized lookup tables that are used by the algorithm, and some more that I could not use because of stack limitation, but still might be usefull for other parts.

This is the comparisson between the current sha256 algo and the new one.
| Algorithm | Current | New | Gain |
|---|---|--|--|
| sha256(32) | 516K | 344K | 33% |
| sha256(80) | 1043K | 759K | 27% |


RobinLinus:
Great improvements! Thank you very much

PR #68, opened 2024-05-17
[On Proving Pairings 2/3] verify pairing with c and wi 

PayneJoe:
Hi, I implemented the Algorithm 9 of [On Proving Pairings](https://eprint.iacr.org/2024/640.pdf) and necessary tests passed:

- [prover-side] compute the proof, denoted as `c` and `wi`, for verifying final exponentiation 
- [verifier-side] compute `c^lambda` within miller loop along with scaled line evaluation `f * wi`
- assert `f * wi * c_inv^lambda = c^{p^3}`

mainly two contributions:
- `bn254::pairing::dual_miller_loop_with_c_wi`, script of verifying dual pairings with `c` and `wi`
- `fflonk::checkpairing_with_c_wi`, test code of verifying dual pairings with `c` and `wi`


RobinLinus:
Impressive work! Thanks a lot

PR #71, opened 2024-05-30
[groth16's verifier] four pairings with one non-fixed on G2

zulunetwork-dev:
Hi, 

We've implemented the groth16's verifier based on [On Proving Pairings](https://eprint.iacr.org/2024/640.pdf), adding support for the non-fixed on G2 after the work of #68.  And the relevant tests already passed.

There are four pairings in groth16's verifier, where three of them are fixed, and only one is **non-fixed**.

Our main contributions:
- `add_line_with_flag`, `double_line` in `pairing.rs`, scripts of the non-fixed line functions
- `quad_miller_loop_with_c_wi` in `pairing.rs`, script of  the quadruple pairing verification
- `msm.rs`, script of msm during preprocessing 
- `groth16` folder, groth16's verifier wrapped with `quad_miller_loop_with_c_wi`

Summary of script size:
- preprocessing part (1-point of MSM) is approximately 2G
- pairing part is approximately 5G
- total size is approximately 7G

The public input length is 1 (1-point MSM). The maximum stack height is approximately 1112. Note that, for the purpose of testation we use the height-free library [rust-bitcoin-scriptexec](https://github.com/BitVM/rust-bitcoin-scriptexec) with a optional tag.

Our team members including @(SuccinctPaul)[https://github.com/SuccinctPaul], @(peinlcy)[https://github.com/peinlcy], our team leader @(Cyimon)[https://github.com/Cyimon], and @PayneJoe . We would especially like to thank https://eprint.iacr.org/2024/640.pdf).


zulunetwork-dev:
the total time cost would be like 50 minutes

src/groth16/offchain_checker.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// refer from Algorithm 5 of "On Proving Pairings"(https://eprint.iacr.org/2024/640.pdf)
+pub fn compute_c_wi(f: ark_bn254::Fq12) -> (ark_bn254::Fq12, ark_bn254::Fq12) {
+    let p = BigUint::from_str_radix(Fq::MODULUS, 16).unwrap();
+    let r = BigUint::from_str(
+        "21888242871839275222246405745257275088548364400416034343698204186575808495617",
+    )
+    .unwrap();
+    let lambda = BigUint::from_str(
```
rot13maxi, line 67: you've already got lambda as a static Lazy in `src/groth16/constants.rs`. might want to grab it from there instead of re-parsing it everytime. consider moving other constants there as well

zulunetwork-dev: yes, that would be better, if time permitted. thanks-)

RobinLinus: @rot13maxi is part of a group of token scammers who are trying to sabotage the BitVM project.

PR #73, opened 2024-06-06
add batched_scalar_mul and reduce cost of fflonk verifier

wz14:
This PR reduces the cost of fflonk verifier from `7456022084` bytes (7.4Gb) to `5649799319` bytes (5.6Gb) and the max stack items number still is 905.

A new function `batched_scalar_mul` for multiple point multiplication is added to G1Projective, which calculates $xA+yB+zC+...$ simultaneously.
Each scalar multiplication currently costs ~0.95Gb, while the function `batched_scalar_mul` costs ~1.5Gb no matter how many items it has.

A example is $xA + yB$. The function first pre-compute some bases $\\{A+B, A, B, 0\\}$, then combine each bit of $x$ and $y$ to a list $\\{0, 1, 2, 3\\}^{254}$. Then, apply double-add to  the 2bit list with these bases.

For fflonk verifier, there are 5 scalar multiplications. This PR uses `batched_scalar_mul<2>` and `batched_scalar_mul<3>` to reduce the cost of fflonk verifier, which will be improved by ~(0.95\*5 - 1.5\*2 = 1.75Gb).

Can fflonk verifier use `batched_scalar_mul<5>` to reduce more? Because of the pre-compute, the bases will cost 2^5 * 27 = 864 stack items and the stack will not be enough.

> This work is inspired by @weikengchen and from bitlayer team, which is a Bitcoin Layer 2 based on the BitVM paradigm.

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -311,6 +311,103 @@ impl G1Projective {
         )
     }
 
+    /// Convert a number to digits
+    fn to_digits_helper<const DIGIT_COUNT: usize>(mut number: u32) -> [u8; DIGIT_COUNT] {
+        let mut digits: [u8; DIGIT_COUNT] = [0; DIGIT_COUNT];
+        for i in 0..DIGIT_COUNT {
+            let digit = number % 2;
+            number = (number - digit) / 2;
+            digits[i] = digit as u8;
+        }
+        digits
+    }
+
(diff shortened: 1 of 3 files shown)


weikengchen:
Wow! This is great.

RobinLinus:
great work!

PR #81, opened 2024-07-18
Remove one g1 scalar mul from msm if the first scalar is 1

ekrembal:
This PR reduces the size of Groth16 verifier from 4B to 3.5B

This contribution is from Citrea Team mainly by @Hakkush-07 
We plan to open-source one more important step in upcoming days, mainly related to splitting the verifier script into 4mb pieces. This PR helps us with that improvement.


ekrembal:
Closing since #83 already implements this.

PR #85, opened 2024-08-07
Input stack optimized Winternitz OTS

bbist:
This PR introduces a input stack optimized version of General Winternitz One Time Signatures.

This work reduces input stack requirement by 55% by while increasing the script size by 65%. The stack size reduction is significantly more valuable that the marginal increase in script size because it allows passing more data do the BitVM script chunk.

The script size increase is considered marginal, because the `stack : script` limits ratio is `1 : 400` for standard transaction. It is `1 : 4192` if we consider entire available block space. The stack optimized version has lower stack utilization per unit of script than the un-optimized version.

This also fixes a minor issue, I think, that resolves the public key conflict with the digit 15.


bbist:
Nevermind. Ended up re-implementing what already exists.

PR #87, opened 2024-09-02
Optimized Fq multiplication using hints with w-width windowed method

bbist:
This PR adds source code for the optimized field multiplication in Bitcoin Script for Bn254 curve used in SNARK verification (i.e. Groth16).

No changes other than additional functions and macros needed for the implementation of the field multiplication. All the scripts that use field multiplications need to be modified to take a hint in input for each of the field multiplication being performed. This implementation also supports linearly combining field multiplications. 

This PR ports the [original implementation](https://github.com/alpenlabs/BitVM/blob/feature/op-tmul-mod/src/bn254/fp254_lc_tmul.rs), to the current BitVM source code using macros instead of advanced rust nightly features.

Please review and give us feedbacks on how we can improve this implementation.

Reference: https://www.alpenlabs.io/blog/releasing-tmul

### DIFF
--- a/src/bn254/fq.rs
+++ b/src/bn254/fq.rs
@@ -1,6 +1,10 @@
+use num_bigint::BigInt;
+use num_traits::{FromPrimitive, Num, ToPrimitive};
 
+use crate::bigint::BigIntImpl;
 use crate::bn254::fp254impl::Fp254Impl;
-
+use crate::pseudo::NMUL;
+use crate::treepp::*;
 
 pub struct Fq;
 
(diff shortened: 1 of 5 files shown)


Hakkush-07:
Hi, thanks for this great improvement. We have been experimenting with tmul and we think we have found a minor bug. When `q=0`, that is when `a*b` is less than the prime modulus, it fails. We think that is because of the `is_positive` function in `bigint/std.rs`. This function assumes 0 is positive while it should not. Below is the fixed version.
```rust
pub fn is_positive(depth: u32) -> Script {
    script! {
        { Self::is_zero_keep_element(depth) } OP_NOT
        { (1 + depth) * Self::N_LIMBS } OP_PICK
        { Self::HEAD_OFFSET >> 1 }
        OP_LESSTHAN OP_BOOLAND
    }
}
```

bbist:
Good catch. Thanks for the response.

Do you think it would make sense to use `is_negative()` instead and do `OP_NOT` or `OP_NOTIF` afterwards?

Or we could use BigInt comparision where `is_positive()` is used. I think this check only happens once, therefore would not add much to the script.

Hakkush-07:
Yes, using `is_negative()` and `OP_NOTIF` should be the best solution. `is_positive()` can stay the same if in this context being "positive" usually means being non-negative. Maybe the name can be changed to `is_nonnegative()` or a small comment can be added.

src/bn254/fq.rs
```diff
@@ -34,6 +38,273 @@ impl Fp254Impl for Fq {
...
+                        script! {
+                            for _ in 1..1<<window {
+                                { T::drop() }
+                            }
+                        }
+                    }
+
+                    // Get modulus window at given index
+                    fn mod_window(index: u32) -> u32 {
+                        let shift_by = MOD_WIDTH * (N_WINDOW - index - 1);
```
Hakkush-07, line 137: I think instead of `const N_WINDOW: u32 = MAIN_LOOP_END / VAR_WIDTH;`, another const `MAIN_LOOP_END / MOD_WIDTH` should be used here for `N_WINDOW`. It does not make a difference when MOD_WIDTH and VAR_WIDTH are the same (I know the best option is when they are equal at 4) but (4, 3) and (3, 4) are not working currently.

bbist: Yeah, moving the N_WINDOW into respective functions makes sense. For mod_window, it should be MAIN_LOOP_END / MOD_WIDTH, and for var_window_script, it should be MAIN_LOOP_END / VAR_WIDTH.

PR #89, opened 2024-09-10
~50% Optimization of Field Multiplication with Lookup Table

fiamma-dev:
We are excited to merge our new optimization on Field Multiplication operations into the BitVM repository, inspired by a paper from Alpen Labs. Key improvements include:

- Reducing Field Multiplication operations from ~140K to ~76K;
- Decreasing the script size for Groth16 by ~1.08GB and for Fflonk by ~0.7GB.

Please note, this new code is not compatible with the current version due to the need to precompute a hint. The changes to the script are as follows:

- Current script code: a, b, Fq_mul;
- New script code: a, b, q, Fq_mul_bucket;

This optimization is particularly suitable for fragmentation. Special thanks to Payne, Cyimon, Paul, and Alpen Labs for their contributions.

### DIFF
--- a/src/bn254/fp254impl.rs
+++ b/src/bn254/fp254impl.rs
@@ -3,13 +3,16 @@ use crate::bigint::bits::limb_to_be_bits;
 use crate::bigint::sub::limb_sub_borrow;
 use crate::bigint::U254;
 use crate::bigint::u29x9::{u29x9_mul_karazuba, u29x9_mul_karazuba_imm, u29x9_mulhi_karazuba_imm, u29x9_mullo_karazuba_imm, u29x9_square};
+use crate::bn254::fq::Fq;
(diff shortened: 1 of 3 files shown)


lucidLuckylee:
Hey thank you for the PR! I have a few questions:
- What is the difference between your PR and the implementation of #87 as supposedly they are based on the same paper?
- Do you have numbers on how many inputs/stack elements we will need for the hints in groth16 or fflonk?

Cyimon:
1. Great, we weren't aware that they merged their implementation as well. 
2. For Groth16, there are 14474 field multiplications in the whole process, so it needs an additional 14474 Fq elements. But for each subscript, it depends on how many field multiplication operations in it

lucidLuckylee (approved):

PR #92, opened 2024-09-13
Schnorr signature verification

chickenblock:
## Implemented

 * schnorr signature verification logic using BitVM's BN254 curve in Bitcoin scripts, and utilities for supportive off-chain signing and verification logic.
 * a small fix to allow G1Affine::into_projective() to also work with identity/zero points.

**to be squash merged.**

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -711,7 +711,20 @@ impl G1Affine {
     }
     // Input Stack: [x,y]
     // Output Stack: [x,y,z] (z=1)
-    pub fn into_projective() -> Script { script!({ Fq::push_one() }) }
+    //pub fn into_projective() -> Script { script!({ Fq::push_one() }) }
+    pub fn into_projective() -> Script {
+        script!{
+            { Fq::is_zero_keep_element(0) }
+            OP_TOALTSTACK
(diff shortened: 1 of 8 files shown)


lucidLuckylee:
Hi @chickenblock!
Thank you for the PR. Although we considered using this BitVM repository as a library for all sorts of scripts that make use of the BitVM paradigm we came to the conclusion that it would just become too much to maintain code that will not be part of the zk verifier script.
Therefore, I would want to only merge commit 1225aac that fixes the `into_projective` script. Do you want to create an extra PR for that or shall I cherrypick the commit?

chickenblock:
> Hi @chickenblock! Thank you for the PR. Although we considered using this BitVM repository as a library for all sorts of scripts that make use of the BitVM paradigm we came to the conclusion that it would just become too much to maintain code that will not be part of the zk verifier script. Therefore, I would want to only merge commit [1225aac](https://github.com/BitVM/BitVM/commit/1225aaccb8bcde8c660a054f59c8363abab17593) that fixes the `into_projective` script. Do you want to create an extra PR for that or shall I cherrypick the commit?

Hi,
You may cherry-pick any of the commits, that are useful for BitVM.

Issue #110, opened 2024-10-07
Optimization: MSM in affine co-ordinates


lucidLuckylee:

lucidLuckylee:
Closed with #114.

PR #114, opened 2024-10-11
Affine versioned MSM 

PayneJoe:
Hi, I replaced MSM with projective coordinate with affine one upon my last PR [#83](https://github.com/BitVM/BitVM/pull/83) , I think I had much underestimated affine's impact on MSM, the result's still beautiful:
- single scalar_mul, script size is reduced by 267328930 bytes
- groth16 verifier with one public input, script size is reduced by 277271131 bytes, from 2652277658 bytes to 2375006527 bytes

Give credit to the authors of [On Proving Pairings](https://eprint.iacr.org/2024/640.pdf), thanks for the backend of [Bitlayer Labs]( https://github.com/bitlayer-org), and thanks for Lukas's pointing out https://github.com/BitVM/BitVM/issues/110.

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1387,12 +1387,7 @@ impl G1Affine {
                 let double_loop = script! {
                     // check before usage
                     { G1Affine::push(*step) }
-                    { Fq::copy(3) }
-                    { Fq::roll(3) }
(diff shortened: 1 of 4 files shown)


lucidLuckylee:
Great work, thank you!

PR #115, opened 2024-10-11
fix G2 element subgroup membership check

dylanCai9:
Hi, I fix the bug mentioned by BitVM /issues/109, support G2 element subgroup membership check and the script len is about 990983 bytes.
Thanks for the backend of [Bitlayer Labs](https://github.com/bitlayer-org).

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1,5 +1,6 @@
 use ark_ec::{AffineRepr, CurveGroup};
 use ark_ff::{AdditiveGroup, BigInteger, Field, PrimeField};
+use bitcoin::opcodes::all::{OP_BOOLAND, OP_FROMALTSTACK, OP_TOALTSTACK};
 use num_bigint::BigUint;
 
 use crate::bigint::U254;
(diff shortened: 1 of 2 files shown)


lucidLuckylee:
What Bitlayer Backend?

dylanCai9:
> What Bitlayer Backend?
I am an contributor from bitlayer lab.

lucidLuckylee:
Ah, that's what it meant. Thank you and the Bitlayer team!

manishbista28:
Hi -- the implementation here seems to be for Point-On-Curve check and not subgroup membership check. From what I understand, both of these checks are essential and the issue raised was for the latter check.
https://github.com/Consensys/gnark-crypto/blob/11d2ed41a3d3470f9cf21a8e1cd4ca5d60a3983f/ecc/bn254/g2.go#L491

@dylanCai9, @wz14 -- please correct me if I'm wrong.
Thanks :)

dylanCai9:
> Hi -- the implementation here seems to be for Point-On-Curve check and not subgroup membership check. From what I understand, both of these checks are essential and the issue raised was for the latter check. https://github.com/Consensys/gnark-crypto/blob/11d2ed41a3d3470f9cf21a8e1cd4ca5d60a3983f/ecc/bn254/g2.go#L491
> 
> @dylanCai9, @wz14 -- please correct me if I'm wrong. Thanks :)

Oh, I missed the additional check for the G2 subgroup, mistakenly assuming that verifying it the same way as for the G1 subgroup in BN254 would be sufficient. Thank you for pointing out this issue!

manishbista28:
no worries, very happy to be of any help :)

PR #116, opened 2024-10-11
Optimize Fq2 mul (hinted) to use 2 lc2 instead of 3 lc1 tmuls & Fix Fq neg

### LINKED ISSUE #111: Optimization: fp2 should use 2 lc2's instead of 3 lc1's (Karazuba).

Hakkush-07:
This PR optimizes Fq2::hinted_mul from 205.715 bytes to 190.871 bytes by using 2 lc2 tmuls instead of 3 lc1 tmuls. This reduces hinted groth16 verifier from 1.33GB to 1.26GB. Also, stack usage is improved, for example Fq12::hinted_mul now uses less than 1000 stack elements (882, was 1080 before).

Also, Fq::neg function was thinking negative of zero to be prime modulus. Now, it correctly gives zero.

Closes #111

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1596,7 +1596,7 @@ mod test {
                 OP_TRUE
             };
             println!("curves::test_hinted_add = {} bytes", script.len());
-            run(script);
+            assert!(execute_script(script).success);
         }
     }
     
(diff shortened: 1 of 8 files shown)


lucidLuckylee:
Thank you, great work!

PR #117, opened 2024-10-12
Hint msm affine

dylanCai9:
hi, I supported the hinted_msm_affine_version groth16 verifier based on https://github.com/BitVM/BitVM/pull/114  and https://github.com/BitVM/BitVM/pull/93, and the hinted_groth16_verifier script len is 1127900767bytes (about 1.05G). This reduces hinted groth16 verifier from 1.26GB to 1.05GB. Thanks @PayneJoe , @wz14, @cyl19970726 for your help and the support from bitlayer labs.

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1286,8 +1286,8 @@ impl G1Affine {
         }
     }
 
-    pub fn hinted_check_line_through_point(point: ark_bn254::G1Affine, c3: ark_bn254::Fq, c4: ark_bn254::Fq) -> (Script, Vec<Hint>) {
-        let (hinted_script1, hint1) = Fq::hinted_mul_by_constant(point.x, &c3);
+    pub fn hinted_check_line_through_point(x: ark_bn254::Fq, c3: ark_bn254::Fq, c4: ark_bn254::Fq) -> (Script, Vec<Hint>) {
+        let (hinted_script1, hint1) = Fq::hinted_mul_by_constant(x, &c3);
         let script_lines = vec![
             //x y
             Fq::roll(1),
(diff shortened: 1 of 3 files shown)


src/bn254/curves.rs
```diff
@@ -1447,6 +1528,120 @@ impl G1Affine {
         }
     }
 
+    pub fn hinted_scalar_mul_by_constant_g1(
```
PayneJoe, line 1531: should it be hinted_scalar_mul_by_constant_g1_affine?

PayneJoe:
great!

lucidLuckylee:
Thank you for the PR!

manishbista28:
Hi -- I have a questions on this implementation.
While collecting line coefficients for doubling operation in msm::affine_double_line_coeff(), you seem to have done sth like
T <- T + [2^w -1]T instead of repeated doublings, thus returning add_line coefficients of chord through points (T, [2^w-1]T). 
My question is: how is it ensured based on the returned add_line coefficients that exactly the required number of doublings (w) was done on point T ? I mean the coefficient could have been for a chord through arbitrary points (T, T').

PayneJoe:
> Hi -- I have a question on this implementation. While collecting line coefficients for doubling operation in msm::affine_double_line_coeff(), you seem to have done sth like T <- T + [2^w -1]T instead of repeated doublings, thus returning a single add_line coefficients of chord through points (T, [2^w-1]T). My question is: how is it ensured based on the returned add_line coefficients that exactly the required number of doublings (w) was done on point T ? I mean the coefficient could have been for a chord through arbitrary points (T, T').
> 
> I thought one would need to return 'w' many doubling coefficients to show that doublings were done properly.

If you refer the original MSM implementation, you will notice that the right operator of addition OP is obtained with the help of a predefined lookup table, and the left operator of addition OP is a w-window period doubling and it's a constant point after fixing the window size on setup phase (there's no need to do w-time's doubling to get the constant point).  In a conclusion, this is just a script version (translation) of rust/go versioned MSM.

PR #121, opened 2024-10-24
feat: optimize winternitz

sander2:
Using 160-bit messages, this reduces the script size from 26.9375 bytes / bit to 26.1125 bytes / bit. (saves 3 ops per digit)

The optimization only works for 4-bit digits and up. For now it's implemented only for 4-bit digits. See the comments for details

(also changed `run` to `execute_script` since `run` doesn't work even in `main`)


lucidLuckylee:
That is great! Thank you!

PR #123, opened 2024-10-29
feat: Add helper functions to use signatures as witness

archidoge0:
Changelogs:
- Add functions with the suffix `_witness` to use the signature as witness data.
- Format the code in the `signatures` folder.


lucidLuckylee:
Thank you!

PR #124, opened 2024-10-30
change algebra dependency to the version with new ATE_LOOP_COUNT

Hakkush-07:
This PR changes the algebra dependency to include the new ATE_LOOP_COUNT change, while keeping the work done by by Antalpha-Labs. With this change hinted Groth16 verifier becomes 1.002 GB instead of 1.05 GB.


lucidLuckylee:
Thank you guys!

PayneJoe:
So the representation of constant parameter of ate pairing miller loop `\lambda` in Arkworks is not optimal (NAF)? Good catch! It seems that Arkworks codebase still needs some improvement, and maybe we also need some check before using it as benchmark-)

PR #128, opened 2024-11-06
optimized sha256_u4_stack script length and stack size

TomerStarkware:


lucidLuckylee:
Hi thanks for the PR!
I know you rely on the bitcoin stack lib but we moved on from the old script macro (the one that uses `define_pushable!()`) in the repo. The script lib and sha256_u4 scripts have to be updated to use that new macro but I didn't have time to do that and I don't think I will any time soon.
Appreciate the optimizations and I'll leave this open until I, you or someone else has refactored it to the compile-based macro version.

stillsaiko:
merged on pr [#132](https://github.com/BitVM/BitVM/pull/132) closing

PR #131, opened 2024-11-08
Add CLI and Winternitz Commit Logic to Client

justin-elementlabs:


justin-elementlabs:
The current tests for Winternitz are not running in the BitVM:main repo and continue to fail in this PR as well.
`cargo test -- --nocapture winternitz`

In our repo (https://github.com/elementlabs42/BitVM-playground/tree/rc), the tests above run fine. It seems there's a bug or issue with:

rust-bitcoin-script-0226d06c574a0653/9c6074d/src/analyzer.rs:259:25:
assertion `left == right` failed: only_if_flow shouldn't change altstack status StackStatus { deepest_stack_accessed: 0, stack_changed: 0, deepest_altstack_accessed: 0, altstack_changed: 1 }
	at pos 1791
	in "for bitvm::signatures::winternitz_compact::checksig_verify"
  left: 1
 right: 0

archidoge0:
@justin-elementlabs @lucidLuckylee 
Hi there! I hope you're doing well. I'm reaching out to inquire about the current status of the `kick_off_2` transaction. As widely discussed, an `assert` transaction should include all bit commitments of hashes for the input of each chunk.

Based on my tests, the minimal implementation from AlpenLabs still involves verifying around 27k hashes in the `kick_off_2` -> `assert` transaction pair. Therefore, an additional 30 input -> output entries in the pair appear necessary to avoid hitting the stack limit.

Could you please share your thoughts on this, given your proposal regarding the architecture of connectors?

justin-elementlabs:
We've updated the Winternitz lib so that Winternitz tests are passing now:
`cargo test -- --nocapture winternitz`

justin-elementlabs:
> @justin-elementlabs @lucidLuckylee Hi there! I hope you're doing well. I'm reaching out to inquire about the current status of the `kick_off_2` transaction. As widely discussed, an `assert` transaction should include all bit commitments of hashes for the input of each chunk.
> 
> Based on my tests, the minimal implementation from AlpenLabs still involves verifying around 27k hashes in the `kick_off_2` -> `assert` transaction pair. Therefore, an additional 30 input -> output entries in the pair appear necessary to avoid hitting the stack limit.
> 
> Could you please share your thoughts on this, given your proposal regarding the architecture of connectors?

We haven't incorporated the chunked verifier changes to the graph, yet. We'll probably look at it in about 3 weeks.

lucidLuckylee:
I opened #134 to merge main into this and update the Winternitz specific parts.

src/bridge/connectors/connector_6.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn generate_taproot_leaf_0_script(&self) -> ScriptBuf {
+        let destination_network_txid_public_key = self.commitment_public_keys
+            [&CommitmentMessageId::PegOutTxIdDestinationNetwork]
+            .get_standard_variant_ref();
+        let source_network_txid_public_key = self.commitment_public_keys
+            [&CommitmentMessageId::PegOutTxIdSourceNetwork]
+            .get_standard_variant_ref();
+
+        script! {
+          { check_hash_sig(&PublicKey::from(destination_network_txid_public_key), DESTINATION_NETWORK_TXID_LENGTH_IN_DIGITS) }
```
lucidLuckylee, line 53: Are you sure you want to use `check_hash_sig` for the txids? If they are 32 bytes it should be cheaper to use them as the message directly instead of hashing them with the `blake3` script. If you want to use `check_hash_sig` then you would have to provide the message in bytes not digits.

lucidLuckylee: Just compared that in a test and using a Winternitz signature directly uses about 10k opcodes while only hashing 64 Bytes is already 100k.

PR #139, opened 2024-11-29
Affine infinity fix

### LINKED ISSUE #137: Proofs with Bad Scalar format in MSM
**What does not work:** Groth16 verifier for some proofs (native, hinted, chunked), namely for proofs with a _bad_ public input.

**Bad Scalar:** If the 12-bit chunk representation of a scalar (public input) contains a 0 (zero) chunk, it fails. For example, that includes small public inputs starting with a few zero bits.

**The Reason:** The functions calculating alpha and bias, affine_double_line_coeff and affine_add_line_coeff, work with non-infinity curve points. However, infinity points are provided as inputs to these function in case of a bad scalar.

Affine curve points have x and y coordinates. Infinity should also have a representation, and it is given (x, y) = (0, 0), which does not actually correspond to a valid point.

- affine_add_line_coeff(infinity, infinity): attempts to divide by 0, fails at script generation stage.
- affine_add_line_coeff(infinity, a): works but gives wrong results because of the incorrect numerical representation of infinity. Script is generated, but fails with correct inputs.
- affine_double_line_coeff(infinity): attempts to add infinity to infinity, fails at script generation stage.

**A Possible Solution:** Adding if checks to both rust calculation part and scripts.

Hakkush-07:
This PR fixes the issue with affine infinity as the inputs of affine_add and affine_double functions. Also, T+(2^w-1)T logic is changed to w times doubling in G1 scalar multiplication. Slightly (around %6) increases the script size. This error occurs mainly for small public inputs so some tests covering this are added. Closes #137

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1534,14 +1534,7 @@ impl G1Affine {
                     let point_after_double = trace_iter.next().unwrap();
                     let double_loop = script! {
                         // check before usage
-                        { G1Affine::is_zero_keep_element() }
-                        OP_NOTIF
-                            { G1Affine::check_double(double_coeff.0, double_coeff.1) }
-                            // FOR DEBUG
-                            // { G1Affine::push(point_after_double.clone()) }
-                            // { G1Affine::equalverify() }
-                            // { G1Affine::push(point_after_double.clone()) }
-                        OP_ENDIF
+                        { G1Affine::check_double(double_coeff.0, double_coeff.1) }
                     };
                     loop_scripts.push(double_loop.clone());
                 }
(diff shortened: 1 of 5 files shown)


src/bn254/curves.rs
```diff
@@ -1467,19 +1528,16 @@ impl G1Affine {
...
-                };
-                loop_scripts.push(double_loop.clone());
+                for _ in 0..depth {
+                    let double_coeff = coeff_iter.next().unwrap();
+                    let step = step_p_iter.next().unwrap();
+                    let point_after_double = trace_iter.next().unwrap();
+                    let double_loop = script! {
+                        // check before usage
+                        { G1Affine::check_double(double_coeff.0, double_coeff.1) }
+                    };
```
lucidLuckylee, line 1538: Couldn't this be
```
let double_loop = G1Affine::check_double(double_coeff.0, double_coeff.1);
```

src/chunker/chunk_scalar_mul.rs
```diff
@@ -49,31 +49,129 @@ pub fn chunk_hinted_scalar_mul_by_constant<T: BCAssigner>(
...
+                let segment = Segment::new_with_name(format!("{}_loop_{}_piece_1", prefix, i), segment_script)
+                    .add_parameter(&type_acc)
+                    .add_result(&update)
+                    .add_hint(hints.clone());
+                hints.clear();
+                segments.push(segment);
+
+                type_acc = update;
+
+                for _ in (i_step / 3)..(2 * i_step / 3) {
```
manishbista28, line 84: could there be a case where depth > (i_step/3) but less than (2* i_step / 3) ? in that case it seems like execution of this loop will not be necessary.
for i_step = 12, it will not be the case, as the final loop has depth = 2 (=254 - 12n) so control won't flow to this branch so is not a problem.

Hakkush-07: yes you are right. I knew it will not be problem for i_step=12 as you indicated but better include all possible cases properly. done.

lucidLuckylee:
Thank you for looking into the issue!

PR #140, opened 2024-12-02
fix hint for pairing result

freddi8e:
Current implementation needs to compare accumulator with c^(q^3), it doesn't follow the algorithm 9 in "on proving pairing" and give attacker a chance to put a fake hint.
Special thanks to @wz14 .  This work is from bitlayer.org.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.20"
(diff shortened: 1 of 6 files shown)


PayneJoe:
Good catch! Actually the original `hint` equals `c_inv^{p^3}`, it's a variant of witness `c`, and it is computed by verifier itself not provided by prover. So I should have forgotten putting `c_inv^{p^3}` into script, it's a bug on verifier script.

PR #142, opened 2024-12-03
Disprove execution

wz14:
This PR offers a function to find which chunk is available when the operator did some malicious thing. This function re-runs chunker using the wrong proof to get a wrong-proof assigner, then checks the inconsistency between the wrong-proof assigner and the assigner recovered by assert transaction.

There are two cases a malicious operator can do.
-  Operator put a wrong proof to generate all intermediate values, say $$S$$. In this case, the final accumulator will be wrong, and the challenger can take collateral by executing the "accumulator verifier" chunk. This case is covered by unit test `test_wrong_proof`.
-  Operator put a wrong proof and modify some intermediate values to make some chunks pass, say $$S'$$. In this case, there must be a chunk whose `parameter` is the same as in $$S$$, but the `result` is different in $$S$$. The challenger can execute this chunk. This case is covered by unit test `test_wrong_proof_and_modify_intermediates`.

### DIFF
--- a/src/chunker/assigner.rs
+++ b/src/chunker/assigner.rs
@@ -64,6 +64,11 @@ impl BCAssigner for DummyAssinger {
         &self,
         elements: BTreeMap<String, Rc<Box<dyn ElementTrait>>>,
     ) -> Vec<Vec<RawWitness>> {
+        for (key, _) in self.bc_map.iter() {
+            if !elements.contains_key(key) {
+                println!("unconsistent key: {}", key)
+            }
+        }
         assert_eq!(elements.len(), self.bc_map.len());
         vec![elements
             .iter()
--- a/src/chunker/assigner.rs
+++ b/src/chunker/assigner.rs
@@ -17,8 +17,10 @@ pub trait BCAssigner: Default {
         elements: BTreeMap<String, Rc<Box<dyn ElementTrait>>>,
     ) -> Vec<Vec<RawWitness>>;
     /// recover hashes from witnesses
-    fn recover_from_witness(&self, witnesses: Vec<Vec<RawWitness>>)
-        -> BTreeMap<String, BLAKE3HASH>;
+    fn recover_from_witness(
+        &mut self,
+        witnesses: Vec<Vec<RawWitness>>,
+    ) -> BTreeMap<String, BLAKE3HASH>;
 }
 
 #[derive(Default)]
@@ -28,6 +30,9 @@ pub struct DummyAssinger {
 
 impl BCAssigner for DummyAssinger {
     fn create_hash(&mut self, id: &str) {
+        if self.bc_map.contains_key(id) {
+            panic!("varible name is repeated, check {}", id);
+        }
         self.bc_map.insert(id.to_string(), id.to_string());
     }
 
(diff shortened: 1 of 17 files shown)


src/chunker/assigner.rs
```diff
@@ -1,18 +1,35 @@
...
+pub trait BCAssigner: Default {
     /// check hash
     fn create_hash(&mut self, id: &str);
+    /// return a element of
     fn locking_script<T: ElementTrait + ?Sized>(&self, element: &Box<T>) -> Script;
     fn get_witness<T: ElementTrait + ?Sized>(&self, element: &Box<T>) -> RawWitness;
+    /// output sciprt for all elements, used by assert transaction
+    fn all_intermediate_scripts(&self) -> Vec<Vec<Script>>;
+    /// output witness for all elements, used by assert transaction
+    fn all_intermeidate_witnesses(
```
lucidLuckylee, line 15: typo in `intermediate`

src/chunker/disprove_execution.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    // if all intermediate values is identical, then return the final chunk
+    for (idx, segment) in segments.iter().enumerate() {
+        if segment.is_final() {
+            return Some((idx, segment.witness(assigner)));
+        }
+    }
+
+    panic!("shouldn't be here")
```
lucidLuckylee, line 70: Can you clean up the todo, comments and panic in this function?

lucidLuckylee:
I think if we use `OP_VERIFY` in the script then we have to execute the actual script to check if the operator lied. Otherwise, they can provide a proof that has all correct bit commitments in the execution trace while failing an `OP_VERIFY` check for them.

src/chunker/disprove_execution.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+        let mut is_result_equal = true;
+        for result in segment.result_list.iter_mut() {
+            if result.to_hash().unwrap() != hash_map.get(result.id()).unwrap().clone() {
+                is_result_equal = false;
+                // replace the result to hash_map
+                *result = Rc::new(Box::new(dummy_element(
+                    &result.id(),
+                    hash_map.get(result.id()).unwrap().clone(),
+                )));
```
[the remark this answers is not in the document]
wz14: This function is used when the operator has released all intermediate state hashes, and a segment will first check bit commitment and then run the chunk. So for the inconsistent `result`, it needs to replace all inconsistent results to make sure that the bit commitment check can pass.

The incorrect `param` is useless, because we don't know the behavior of segment with incorrect `param`s. But, we can always find the segment containing consistent `param` and inconsistent `result`.

src/chunker/assigner.rs
wz14: fixed

src/chunker/disprove_execution.rs
wz14: done

wz14:
Fix repeated variable name `{}_loop_{}_piece_1` induced from https://github.com/BitVM/BitVM/pull/139

https://github.com/BitVM/BitVM/blob/c00a487360bbbf5d6e6510a611a43c897504e351/src/chunker/chunk_scalar_mul.rs#L73-L79

src/chunker/disprove_execution.rs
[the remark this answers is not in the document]
lucidLuckylee: I see. So the verifier is supposed to create its own `assigner` by executing the segments without the Winternitz part and retrieve the hashes that way. Then they can check for errors in the bit commited hashes from the operator. Do I understand that correctly?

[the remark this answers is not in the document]
wz14: Yes, you're right. A different type of assigner is needed as a challenger. Running the chunk one by one is another obvious solution, but I think this way is more efficient.

PR #147, opened 2024-12-10
Change G1Affine hinted check functions

Hakkush-07:
This PR changes G1Affine functions hinted_check_double, hinted_check_add and related functions to ensure that no multiple alpha and bias hints are supplied but instead they are copied for usages in multiple functions.


lucidLuckylee:
Thank you! Can you change the tests back to use the `run` function and comment out the stack analyzer part in the `run` function instead. I will later take a look at the stack analyzer and fix it.

PR #153, opened 2024-12-15
feat: add hinted fp12 inverse

manishbista28:
This PR requests addition of hinted Fp12 inverse.
It is useful because pairing operation makes use of auxiliary hint "[c](https://github.com/BitVM/BitVM/blob/main/src/groth16/offchain_checker.rs#L96)" and its inverse.

It seems cheaper to do the inverse directly on-script instead of receiving it as bit commitment input.
Script and max stack size used by a single joint script is  6,168,525 bytes and stack 1116 respectively.
This script can be split into 3 chunks with the intermediates hashed, each of these chunks fit within 4 MB and stack limit.

In total, the cost on Assert Tx is from having to bit commit 3 hashed elements.

Tests have been added and does not impact any other code.

### DIFF
--- a/src/bn254/fq12.rs
+++ b/src/bn254/fq12.rs
@@ -679,56 +679,50 @@ impl Fq12 {
         aux
     }
 
-    pub fn hinted_inv(a: ark_bn254::Fq12) -> (Script, Vec<Hint>) {
-
-        let t1 = a.c1 * a.c1;
-        let t0 = a.c0 * a.c0;
-        let yt1 = Self::mul_fp6_by_nonresidue_in_place(t1);
-        let t0 = t0-yt1;
-        let t1 = t0.inverse().unwrap();
-        let c0 = a.c0 * t1;
-        let c1 = -a.c1 * t1;
-        assert_eq!(ark_bn254::Fq12::new(c0, c1), a.inverse().unwrap());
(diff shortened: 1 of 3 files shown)


manishbista28:
chunked version of fp12 inverse is not included here.

lucidLuckylee:
> chunked version of fp12 inverse is not included here.

Are you planning on submitting a PR for a chunked version?

manishbista28:
I have already implemented a chunked version of this which I can include in this PR.
Was wondering how helpful can this direct inversion be to the current chunker ?

@Hakkush-07 , @wz14 do you have answer if it can be useful to you with the chunker ?

freddi8e:
> @wz14 do you have answer if it can be useful to you with the chunker

yes, it is  also useful for chunker , please include the chunked version, thanks.

manishbista28:
I have split the hinted_fp12 inverse into 3 parts, each fits within a tapscript and adjusted [tests](https://github.com/BitVM/BitVM/blob/3e42a6c8bebde84ea54451d87fa3c2594763ce07/src/bn254/fq12.rs#L1200) accordingly.

I haven't integrated it to the chunker implementation though. I did not feel very sure about the changes required but I assume it may not be as difficult to integrate it with the examples given.

freddi8e:
that is great, thank you,

manishbista28:
thanks, in that case, should be good to merge

PR #154, opened 2024-12-15
fix(blake3_u4): issue with insufficient size of precompute table

manishbista28:
Current blake3_u4 implementation has an issue which generates incorrect output hash for some input message.

blake3_u4 operates over nibbles and looks up modulo and quotient values in a precompute table. The size of table depends upon the range of input index. 
For example if you want to sum three number 0xfe + 0xff + 0xff, the impl proceeds in following steps:
1. sum lower limb: 0xe + 0xf + 0xf => 0x2c (=44) 
2. find 44th index in quotient and modulo table to obtain 2 and c respectively. pass 2 as carry.
3. sum higher limb: 0xf + 0xf + 0xf + 0x2 = 0x2f (=47)
4. find 47th index in quotient and modulo table. But 47th index doesn't exist because the [table](https://github.com/BitVM/BitVM/blob/main/src/u4/u4_add_stack.rs#L70) has total size 47 starting from 0. As such any value above this stack position is returned, thus generating wrong result.

The fix adds an extra element on the precompute table making it of size 48. Since blake3_u4 only adds 3 elements at max, this new precompute table length suffices for the entire range of input.

A test has been added which fails if the fix is not present.


jonasmartin:
@manishbista28 Nice catch!
I miss calculated the maximun number possible to be added in a limb. The fix looks good to me.
👍

src/hash/blake3_u4.rs
```diff
@@ -688,6 +688,70 @@ mod tests {
...
+        stack.custom(
+            script! { { inp } },
+            0,
+            false,
+            0,
+            "msg",
+        );
+
+
+        let start = stack.get_script_len();
```
jonasmartin, line 736: The script size should be obtained using `stack.get_script().len()`

lucidLuckylee:
Nice work! Thank you for the PR!

PR #156, opened 2024-12-15
feat(curves): add hinted is_on_curve check

manishbista28:
The G1 and G2 elements of groth16 proof need to be validated to be points on the curve. This PR requests adding hinted version of these scripts.

While non-hinted version of each of these were already present, this version is helpful because of the reduction in script size. For example, test_hinted_g2_affine_is_on_curve uses around 467K script size, while test_g2_affine_is_on_curve uses around 992K.

Unit tests added for each.

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1972,6 +1972,29 @@ impl G1Affine {
         }
     }
 
+    pub fn hinted_is_on_curve(x: ark_bn254::Fq, y: ark_bn254::Fq) -> (Script, Vec<Hint>) {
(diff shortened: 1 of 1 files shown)


manishbista28:
this check is only necessary at the beginning of the pairing algorithm.
G1 and G2 points should be validated to not be point at inf and are points on curve.
current chunker impl doesn't seem to use it.

manishbista28:
i'd suggest merging this as well.
chunker impl can get to it after higher priority [bugs](https://github.com/BitVM/BitVM/pull/163)  are dealt with.

Issue #157, opened 2024-12-16
Chunker: tangent check and line check for no fixed point


wz14:
Almost all elliptic curve operations are computed by the precompute method. There are two cases for fixed point and non-fixed point(q4 in groth16). I think both cases are implemented by putting the coefficients of the precomputed line on the witness (hint). It gives the verifier an opportunity to bypass the chunk.

- For fixed point: The precomputed coefficients should be put in the script.
- For non-fixed point: The precomputed coefficients should be committed by bit commitment.

wz14:
This can be considered as remaining issues of chunker.

lucidLuckylee:
Can you reference the functions in the current codebase?

wz14:
> Can you reference the functions in the current codebase?

Please check this PR https://github.com/BitVM/BitVM/pull/163. @lucidLuckylee

PR #162, opened 2024-12-17
fix: ensure same line coeff in  g2 affine dbl and add

manishbista28:
Fixes issues with separate line coefficients being used in g2 affine point operation functions.
Tests included.

### DIFF
--- a/src/bn254/pairing.rs
+++ b/src/bn254/pairing.rs
@@ -527,7 +527,7 @@ impl Pairing {
                 fx.mul_by_034(&coeffs.0, &c1new, &c2new);
 
                 let (hinted_script, hint) =
-                    hinted_ell_by_constant_affine(f, -p.x / p.y, p.y.inverse().unwrap(), coeffs);
+                    hinted_ell_by_constant_affine_and_sparse_mul(f, -p.x / p.y, p.y.inverse().unwrap(), coeffs);
                 scripts.push(hinted_script);
                 hints.extend(hint);
                 f = fx;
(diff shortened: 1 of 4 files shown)


manishbista28:
Remaining:
Apply necessary changes to test_hinted_quad_miller_loop_with_c_wi.
Left because the test hasn't been functional for some time now.

manishbista28:
PR in draft because awaiting discussion before merge.

manishbista28:
requesting bitlayer team @wz14  to review changes to chunk_non_fixed_point.rs specifically.
the tests in that file, assertion and disprove is working though.

freddi8e:
> requesting bitlayer team @wz14 to review changes to chunk_non_fixed_point.rs specifically. the tests in that file, assertion and disprove is working though.

It seems to solve same issue as   [issue 157](https://github.com/BitVM/BitVM/issues/157).
Actually, we are fixing the issue 157  and will finish soon;  in the fix,  the non-fixed point  coefficients is committed by bit commitment,  which can  ensure  same value is used by check_q4 and accumulator.

manishbista28:
@freddi8e 
> Actually, we are fixing the issue 157 and will finish soon; in the fix, the non-fixed point coefficients is committed by bit commitment, which can ensure same value is used by check_q4 and accumulator.

i am unsure why the line coefficients belonging to non-fixed point has to be bit-committed. Are the same line coefficients used in two different tapscripts ?

Hakkush-07:
This is extending the work of ensuring same hints in G1Affine to G2Affine. It looks good to me.

PR #163, opened 2024-12-18
fix issue 157

freddi8e:
Fix  [issue 157](https://github.com/BitVM/BitVM/issues/157)

Special thanks to @wz14 . This work is from bitlayer.org.

### DIFF
--- a/bitvm/src/bn254/fp254impl.rs
+++ b/bitvm/src/bn254/fp254impl.rs
@@ -625,35 +625,6 @@ pub trait Fp254Impl {
         (script, hints)
     }
 
-    // TODO: Optimize by using the constant feature
-    fn hinted_mul_by_constant_stable(
-        a: ark_bn254::Fq,
-        constant: &ark_bn254::Fq,
-    ) -> (Script, Vec<Hint>) {
-        let mut hints = Vec::new();
-        let x = BigInt::from_str(&a.to_string()).unwrap();
-        let y = BigInt::from_str(&constant.to_string()).unwrap();
-        let modulus = &Fq::modulus_as_bigint();
-        let q = (x * y) / modulus;
-
-        let script = script! {
-            for _ in 0..Self::N_LIMBS {
-                OP_DEPTH OP_1SUB OP_ROLL // hints
-            }
-            // { fq_push(ark_bn254::Fq::from_str(&q.to_string()).unwrap()) }
-            { Fq::roll(1) }
-            //{ fq_push_not_montgomery(*constant) }
-            for _ in 0..Self::N_LIMBS {
-                OP_DEPTH OP_1SUB OP_ROLL // hints
-            }
-            { Fq::tmul() }
-        };
-        hints.push(Hint::BigIntegerTmulLC1(q));
-        hints.push(Hint::Fq(*constant));
-
-        (script, hints)
-    }
-
     fn hinted_mul_keep_element(
         mut a_depth: u32,
         mut a: ark_bn254::Fq,
--- a/bitvm/src/bn254/fq2.rs
+++ b/bitvm/src/bn254/fq2.rs
@@ -398,39 +398,6 @@ impl Fq2 {
         (script, hints)
     }
 
-    pub fn hinted_mul_by_constant_stable(a: ark_bn254::Fq2, constant: &ark_bn254::Fq2) -> (Script, Vec<Hint>) {
-        let mut hints = Vec::new();
-
-        let (hinted_script1, hint1) = Fq::hinted_mul_by_constant_stable(a.c0, &constant.c0);
-        let (hinted_script2, hint2) = Fq::hinted_mul_by_constant_stable(a.c1, &constant.c1);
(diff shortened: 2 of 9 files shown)


lucidLuckylee:
@manishbista28 Do you know the issue and can review this?

manishbista28:
can't the generic point operation functions that assume c3, c4 on stack be used for fixed Q pairing as well ? there seems to be a _stable and c3,c4 hardcoded function now.

manishbista28:
It seems you have used (c3, c4) across tapscripts i.e. between chunk accumulator (to do line evaluation) and chunk q4 (to do point operation). In that case, I agree that you will have to bitcommit to (c3, c4).

You seem to have added hardcoded function to reduce the amount of line-coeff data to bit commit to. 
My suggestion would be to do all operations that require (c3, c4) within a tapscript, so you don't have to bitcommit to them to being with and as such don't need two versions of the same function.

If that is too much work for now, I'd suggest following changes to the current scope only related to convention and test.

1. that all line operation functions which assume [c3, c4] on stack have the same name as previous i.e. without suffix "_stable". This will make it easier to not have to adapt to that name change everywhere.
2. the implementation of the non-fixed pairing ops seems to assume [c3, c4] on stack which I support. I have added tests for them, could you please include tests when you're done with changes ?
4. Finally could you also ensure other scripts like non-chunked groth16 verifier do not break during this process by running their respective tests in the end ?

freddi8e:
(1) "*_stable" function  is removed;
(2) the following test cases are updated and passed.  
	test_hinted_check_chord_line, 
    test_hinted_check_tangent_line,
	test_hinted_affine_double_line,
	test_hinted_affine_add_line 
(3)  the following test cases passed.  
chunk_groth16_verifier
	chunker
		test_hinted_groth16_verifier_small_public
		test_hinted_groth16_verifier
		test_hinted_groth16_verifier_stable
	disprove
		test_wrong_proof
		test_wrong_proof_and_modify_intermediates
		offchain_check_wrong_proof
non-chunked-groth16
	test_groth16_verifier_native_small_public
	test_hinted_groth16_verifier

manishbista28:
looks good to me, thanks. 
requesting @Hakkush-07 to review as well.

src/bn254/pairing.rs
```diff
@@ -784,6 +784,7 @@ impl Pairing {
                 let q4_new = ark_bn254::G2Affine::new(q4.x * beta_22, q4.y);
 
                 let (hinted_script, hint) = hinted_check_chord_line(
+                    // let (hinted_script, hint) = hinted_check_chord_line(
```
lucidLuckylee, line 787: Let's remove the comment. It doesn't add anything

src/bn254/pairing.rs
```diff
@@ -845,15 +846,46 @@ impl Pairing {
...
+
+                    // -- push c3,c4 to stack  
+                    script_lines.push(fq2_push_not_montgomery(line_coeffs[num_lines - (i + 2)][j][0].1)); 
+                    script_lines.push(fq2_push_not_montgomery(line_coeffs[num_lines - (i + 2)][j][0].2));     
+                    // [...T4(4),T4(4),C3(2),C4(2)]
+                    // -- move t4 to stack top 
+                    script_lines.push(Fq2::roll(6));
+                    script_lines.push(Fq2::roll(6));
+                    // -- [...T4(4),C3(2),C4(2),T4(4)]
                     script_lines.push(scripts_iter.next().unwrap()); // check_tangent_line(line_coeffs[num_lines - (i + 2)][j][0].1, line_coeffs[num_lines - (i + 2)][j][0].2)
```
lucidLuckylee, line 858: I want to get rid of the `script_lines.push()` lines that are still in the code base and rather just use a direct `script{ ... }` block. Could you refactor them with this PR already?

src/bn254/utils.rs
```diff
@@ -261,8 +268,6 @@ pub fn hinted_ell_by_constant_affine_and_sparse_mul(
     for script_line in script_lines {
         script = script.push_script(script_line.compile());
     }
```
lucidLuckylee, line 270: This could be in a `script { }` block aswell.

src/bn254/pairing.rs
Hakkush-07: Just to note that this was our solution to memory failures that occur when working with large scripts. I agree that the usual way is better.

freddi8e: > I want to get rid of the `script_lines.push()` lines that are still in the code base and rather just use a direct `script{ ... }` block. Could you refactor them with this PR already?

I also agree use direct script!{...} is better, but I think it is better to do it in a new PR,  do you think it is okay ?

src/bn254/pairing.rs
freddi8e: fixed.

src/bn254/utils.rs
freddi8e: fixed.

src/bn254/pairing.rs
lucidLuckylee: Yes

PR #164, opened 2024-12-19
Use workspace

ekrembal:
This PR seperates the Bridge and BitVM into seperate packages.


storopoli:
If this get merged. We should tag `0.1.0`. Living on `master`/`main` is too YOLO and having some sort of way to update through tags makes sense.

ekrembal:
Closed with #176

PR #165, opened 2024-12-23
Feat/blake3 u4 input compact representation

mukeshdroid:
The current blake3_u4 has a limitation that only messages of upto 288 bytes could be hashed. blake3_u4_compact overcomes this limitation by packing nibbles into limbs of u32 until they need to be processed. 

Although there is slight increase in the script size, now larger messages (upto 1024 bytes) can be hashed at once leading to savings in both script size and max stack use when computing hash for Fp6 or Fp12 elements.


src/hash/blake3_official_test_vectors.json
AaronFeickert, line 1: Would recommend including a comment in this file giving a link to its source.

src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,414 @@
...
+        for (_, case) in test_vectors.cases.iter().enumerate() {
+            if case.input_len > 0 && case.input_len <= 1024 {
+                assert_eq!(
+                    case.hash[0..64],
+                    test_blake3_compact_giveninputhex(
+                        gen_inputs_with_padding(case.input_len),
+                        case.input_len as u32
+                    )
+                );
+            }
```
AaronFeickert, line 411: I assume it was checked that this conditional is actually triggered, and that relevant equality checks are really being done?

src/bigint/std.rs
```diff
@@ -289,8 +289,146 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
...
+                        }
+                    }
+
+                }}
+            }
+            OP_DROP // drop accumulator
+        }
+    }
+
+    pub fn pack_limbs<const WINDOW : u32>() -> Script {
```
AaronFeickert, line 380: Missing documentation.

src/bigint/std.rs
```diff
@@ -289,8 +289,146 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
             }
         }
     }
+
+    pub fn unpack_limbs<const WINDOW: u32>() -> Script {
```
AaronFeickert, line 293: Missing documentation.

src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,414 @@
...
+
+use crate::bigint::U254;
+
+use crate::hash::blake3_u4::{compress, get_flags_for_block, TablesVars};
+
+// This implementation assumes you have the input is in compact form on the stack.
+// The message must be packed into multiple of (2 * 9) limbs such that it expands a multiple of 128 nibbles 
+// The padding added by user is removed and 0 is added as padding to prevent maliciously or mistaken wrong padding
+
+pub fn blake3_u4_compact(
```
AaronFeickert, line 16: Missing documentation.

AaronFeickert (changes requested):

src/hash/blake3_official_test_vectors.json
mukeshdroid: Would It make sense to add the link as a comment when we import it rather than to this file directly. That would make it easier to verify the integrity of this file by comparing its hash with the one found at https://github.com/BLAKE3-team/BLAKE3/blob/master/test_vectors/test_vectors.json .

src/hash/blake3_u4_compact.rs
mukeshdroid: yes. I have added tests for cases where it should panic.

src/hash/blake3_official_test_vectors.json
AaronFeickert: Good call!

src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    use super::*;
+    use crate::{execute_script, u4::u4_std::u4_hex_to_nibbles};
+
+    fn add_padding(input: String) -> String {
+        let len_bytes = input.len() / 2;
+        let mut res = String::from(input);
+
+        if len_bytes % 64 != 0 {
+            res.push_str(&"0".repeat((64 - (len_bytes % 64)) * 2)); //zero should be added as padding but this is done intentionally to test if blake3_u4_compact can handle wrong padding
```
AaronFeickert, line 220: Isn't this doing correct padding, not incorrect padding as suggested by the comment?

src/hash/blake3_u4_compact.rs
AaronFeickert, line 1: Can you ensure that test module function-level comments appear before the `#[test]` attribute? Apparently the formatter doesn't have a default style for this, but I think it makes intent more clear. And selfishly, my editor makes it [more difficult](https://github.com/zed-industries/zed/issues/22798) to run tests otherwise!

AaronFeickert:
After discussion, a few notes and ideas for this implementation:
- Avoid the use of `U254`-based packing and unpacking functionality. While these work with the full 256 bits we need for BLAKE3 processing, this approach makes the intent unclear and seems very brittle to future changes to `BigIntImpl`.
- Add fuzz testing. This may necessitate different input handling, since the implementation currently panics on invalid input length.
- Refactor (as feasible) to avoid hex/byte conversions. These conversions occur several times and add complexity. While hex representations can make sense for input or output processing, we should stick with byte representations internally as much as possible.
- Check test outputs off the stack. Right now, test vector expected outputs are pushed to the stack as nibbles and compared to the nibbles of the implementation output. This has a bad code smell, since it relies on packing and script functionality to operate as expected. A more robust approach would be to extract the implementation output and check it against the test vector output off the stack.

manishbista28:
looks good to me

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,644 @@
...
+
+    use super::*;
+    use crate::{execute_script, u4::u4_std::u4_hex_to_nibbles};
+
+    fn add_padding(input: String) -> String {
+        let len_bytes = input.len() / 2;
+        let mut res = String::from(input);
+
+        if len_bytes % 64 != 0 {
+            res.push_str(&"0".repeat((64 - (len_bytes % 64)) * 2)); //zero should be added as padding but this is done intentionally to test if blake3_u4_compact can handle wrong padding
```
AaronFeickert, line 366: This comment seems incorrect; you are adding zero values.

AaronFeickert (changes requested):

AaronFeickert:
I noticed a lot of `println!` output during fuzz testing that makes it difficult to monitor the fuzzer's progress. Would recommend either removing/commenting those, or at least making them conditional on a `not(fuzzing)` target.

mukeshdroid:
The  blake3_u4_compact is a wrapper that utilizes a lot of the inner primitives from blake3_u4. The input format however is U256 ints that are packed into 9 limbs instead of nibbles. 

A Fuzz target has also been incorporated and blake3_u4_compact has been fuzz tested. The setup for fuzz can be utilized to fuzz test other parts of the bitvm repo.

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,617 @@
...
+        } 
+    ),
+    0,
+    false,
+    0,
+    "push_msgs"
+    );
+
+    //random bool to decide if we want to use full table or half table.
+    let use_full_table_toogle: bool = random();
```
AaronFeickert, line 283: Recommend making this a parameter for the function so the caller can control it more directly.

fuzz/fuzz_targets/blake3.rs
```diff
@@ -0,0 +1,23 @@
...
+        // we want to confine length of byte array to 1024
+        let size= u.int_in_range(0..=1024)?;
+        let mut bytes = vec![0u8;size];
+        u.fill_buffer(&mut bytes)?;
+        Ok(LimitedBytes(bytes.to_vec()))
+    }
+}
+
+fuzz_target!(|data: LimitedBytes| {
+    test_blake3_compact_givenbyteslice(&data.0);
```
AaronFeickert, line 22: If the test function is updated to take the table flag as a parameter, you can update the `Arbitrary` implementation to include it under the control of the fuzzer.

fuzz/fuzz_targets/blake3.rs
```diff
@@ -0,0 +1,23 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+use arbitrary::{Arbitrary, Unstructured, Result};
+use bitvm::hash::blake3_u4_compact::test_blake3_compact_givenbyteslice;
```
AaronFeickert, line 5: Does this work with `cargo check` for you? Seems like this doesn't play nicely with the target restriction :(

bitvm/Cargo.toml
```diff
@@ -32,8 +32,14 @@ ark-relations.workspace = true
...
 [profile.dev]
 opt-level = 3
 
 [profile.release]
-lto = true
\ No newline at end of file
+lto = true
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(loom)', 'cfg(fuzzing)'] }
```
AaronFeickert, line 45: Nit: missing newline.

bitvm/src/bigint/mod.rs
```diff
@@ -18,3 +18,4 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
 
 pub type U254 = BigIntImpl<254, 29>;
 pub type U64 = BigIntImpl<64, 16>;
+pub type U256 = BigIntImpl<256,29>;
```
AaronFeickert, line 21:
```suggestion
pub type U256 = BigIntImpl<256, 29>;
```

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,617 @@
...
+///     - Corrects any user-provided padding if it is the last block.
+///     - Computes the hash for the block using `compress` while maintaining intermediate states.
+/// 5. Drops intermediate states and finalizes the hash result on the stack.
+///
+/// ## Stack Effects:
+/// - Temporarily uses the alternate stack for intermediate results and hash computation tables.
+/// - Final result is left on the main stack as a BLAKE3 hash value.
+///
+/// ## Panics:
+/// - If `msg_len` is 0 or greater than 1024 bytes, the function panics with an assertion error.
```
AaronFeickert, line 56: This now supports empty input.

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,617 @@
...
+/// - If the input is not a multiple of 18 limbs, or doesn't unpack to a multiple of 128 nibbles.
+/// - If the stack contains elements other than the message.
+///
+/// ## Note:
+/// This function does not currently support the hash of an empty string. Handling for such cases
+/// can be added as a hardcoded return value.
```
AaronFeickert, line 62: This now supports empty input.

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -0,0 +1,617 @@
...
+
+pub fn blake3_u4_compact(
+    stack: &mut StackTracker,
+    mut msg_len: u32,
+    define_var: bool,
+    use_full_tables: bool,
```
AaronFeickert, line 68: These parameters are not documented.

AaronFeickert (changes requested):

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -520,11 +508,18 @@ mod tests {
             for _ in 0..random_size {
                 random_byte_slice.push(rng.gen());
             }
-            test_blake3_compact_giveninputhex(add_padding(gen_random_hex_strs(random_size as u32)), random_size as u32);
-            test_blake3_compact_givenbyteslice(&random_byte_slice);
+
+            // test with full tables
+            test_blake3_compact_giveninputhex(add_padding(gen_random_hex_strs(random_size as u32)), random_size as u32, false);
```
AaronFeickert, line 513: This should pass a `true` flag, no?

bitvm/src/hash/blake3_u4_compact.rs
```diff
@@ -520,11 +508,18 @@ mod tests {
...
+            test_blake3_compact_givenbyteslice(&random_byte_slice, true);
+        
+            // test with full tables
```
AaronFeickert, line 516: Incorrect comment, since this uses half tables.

AaronFeickert (changes requested):

bitvm/src/hash/blake3_u4_compact.rs
mukeshdroid: yes. Good catch. Fixed in 033beee.

bitvm/src/hash/blake3_u4_compact.rs
mukeshdroid: Corrected @ 033beee.

AaronFeickert (approved):

lucidLuckylee:
Thanks for the PR @mukeshdroid and thank you for the review @AaronFeickert! Will you work on integrating this into the chunker?

PR #166, opened 2024-12-24
feat(msm): add hinted w-windowed glv msm

manishbista28:
This PR includes an optimization over the current implementation of hinted msm. It uses [glv msm](https://github.com/arkworks-rs/algebra/blob/master/ec/src/scalar_mul/glv.rs).

Existing implementation has the following benchmarks.
1. msm for single input is split to around 80 bit commitments
2. cost of script size and as such bit-comms scales linearly i.e. 80*n for n public inputs.

This improvement includes:
1. msm for single input is split to around 16 bit comms.
2. cost of script size (or bitcomms) doesn't scale linearly. e.g. 20 bitcomms for 2 public inputs

Tests have been added.

### DIFF
--- a/src/bn254/curves.rs
+++ b/src/bn254/curves.rs
@@ -1,8 +1,11 @@
 use ark_ec::{AffineRepr, CurveGroup};
-use ark_ff::{AdditiveGroup, BigInteger, Field, PrimeField};
-use num_bigint::BigUint;
+use ark_ff::{AdditiveGroup, BigInteger, Field, MontFp, PrimeField};
+use num_bigint::{BigInt, BigUint, Sign};
+use num_traits::{One, Signed};
+
 
 use crate::bigint::U254;
+use crate::bn254;
 use crate::bn254::fp254impl::Fp254Impl;
 use crate::bn254::fq::Fq;
 use crate::bn254::fq2::Fq2;
@@ -10,9 +13,11 @@ use crate::bn254::fr::Fr;
 use crate::bn254::utils::{fq2_push_not_montgomery, fq_push};
 use crate::treepp::{script, Script};
 use std::cmp::min;
+use std::ops::{AddAssign, Div, Neg, Rem};
+use std::str::FromStr;
 use std::sync::OnceLock;
 
-use super::utils::Hint;
+use super::utils::{fr_push_not_montgomery, Hint};
 
 static G1_DOUBLE_PROJECTIVE: OnceLock<Script> = OnceLock::new();
 static G1_NONZERO_ADD_PROJECTIVE: OnceLock<Script> = OnceLock::new();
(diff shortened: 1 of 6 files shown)


manishbista28:
What's not included here is direct integration with existing chunker.
I believe chunk_msm.rs has not made use of direct msm utils present in bn254/curves -- as such this PR will not affect it.

If these changes are to be integrated, I'd recommend going through it over a separate issue (PR).  This way ongoing work on each line doesn't depend on one another.

manishbista28:
requesting @Hakkush-07 @wz14 for review.

manishbista28:
Short intro to glv:
Given a scalar, glv gives a way to break the scalar into two parts, each of bit length half of the original.
k = k0 + lambda * k1 where lambda is constant.
k < Fr::N_BITS,
k0, k1 < Fr::N_BITS/2

In bitcoin script, 
- we pass k0 and k1 alongside k.
- we check that k0 + k1. lambda == k
- we use k0, k1 with msm
- if k had G1Affine base P, k0 has base P and k1 has base phi(P) where phi(P) is point endomorphism of P
- P and phi(P) are both baked into circuit as lookup tables.
- Thus we have twice the number of scalars, half the bit size of these scalars, and their bases are hardcoded in circuit. We run w-windowed double and add algorithm off-the-box.

https://ethresear.ch/t/fake-glv-you-dont-need-an-efficient-endomorphism-to-implement-glv-like-scalar-multiplication-in-snark-circuits/20394

wz14:
Thank you for mentioning me. Looks good to me. 

I am not sure if we could have two versions of algorithm. There are different purposes for a large size script and plenty of smaller chunks. Some scripts work well in the large script but can't be used in chunks. For example, https://github.com/BitVM/BitVM/blob/cb5fe1807d963d60ca19fb33f0fcfdb61cabf2c9/src/bn254/msm.rs#L287-L299. Here we are not supposed to mix hints and input with scripts in the context of chunks because we haven't known the hint and input when pre-signing. 

Certainly, large script also has its functionality. By checking the large script, we can identify the integrity of zkp verifier more easily.

So, I propose that we can keep the algorithm of two versions aligned through PRs. How do you think of this point? @lucidLuckylee @manishbista28 @Hakkush-07

manishbista28:
Thanks for the review and reply @wz14.

_Discussed internally but adding some context if it can be useful to others._

Large script (unchunked-version of groth16 verifier) and chunked version should reuse as much code as possible. Snippet of the code above is from function "hinted_msm_with_constant_bases_affine" which is called only by unchunked version and serves to demonstrate how the chunks returned by core function:  G1Affine::hinted_scalar_mul_by_constant_g1() can be integrated. As such the hints were hard-coded in the above snippet. 

>  Here we are not supposed to mix hints and input with scripts in the context of chunks because we haven't known the hint and input when pre-signing.

I agree that the hints will be passed by chunks as witnesses. 

> So, I propose that we can keep the algorithm of two versions aligned through PRs

I agree that having this version integrated to chunker will be helpful. I'd also like to hear @Hakkush-07 's feedback as i think he was involved in a lot of changes around msm very recently.

Hakkush-07:
Looks good to me, it can be integrated to chunker easily after this is merged. I think we can also made a version of this msm that just returns one script when refactoring the curve utils, and move this one to chunker part.

manishbista28:
clippy error
```
error: this range is empty so it will yield no values
   --> bitvm/src/bn254/fr.rs:251:38
    |
251 |                             for _ in 0..N_LC-1 {
```
where N_LC = 1;

commented the line [here](https://github.com/BitVM/BitVM/pull/166/commits/01c6454476332aa9f6c4b1b168988b75719f7022)
alternatively you could do sth like: #![allow(clippy::reversed_empty_ranges)] at the top of the file.

manishbista28:
requesting @lucidLuckylee to review.

lucidLuckylee:
Thanks @manishbista28!

PR #169, opened 2024-12-30
feat: add splitted blake3_u32

fiamma-dev:
In a production environment, the verification of ZKP typically includes:
1. `Check Hash(Public input) == Hash value`
2. `Verify ZKP(Proof, Hash) == True`
> Note: This requires replacing the hash algorithm with the blake3 hash when generating the proof. Taking the current Risc0 as an example, the sha256 algorithm is used in snark_to_stark. If you want to complete the proof verification, you need to replace it with the blake3 hash.

</br>
BitVM2 already has the script splitting implementation for point 2, and now we are achieving the point 1 by scripting blake_u32 base on #105 .

The bake_u32 script has 3 sub-script, and each size is 101 KB in Test `test_Blake3_split`.


lucidLuckylee:
Hey thank you for the PR! I am not in the loop with the hashing of the public inputs so hope you don't mind some questions.

1. What public inputs have to be hashed? How large are they? 
    Chainway propose this: https://github.com/chainwayxyz/risc0-to-bitvm2?tab=readme-ov-file#goal. Is that what you have in mind too? We don't want to hash the entire journal in bitvm.
2. Why do you need another `blake3_script.rs`?
3. Why does the blake3 script have to be split? It seems feasible to run the entire hashing in < 4MB

Cyimon:
> Hey thank you for the PR! I am not in the loop with the hashing of the public inputs so hope you don't mind some questions.
> 
> 1. What public inputs have to be hashed? How large are they?
>    Chainway proposes this: https://github.com/chainwayxyz/risc0-to-bitvm2?tab=readme-ov-file#goal. Is that what you have in mind too? We don't want to hash the entire journal in bitvm.
> 2. Why do you need another `blake3_script.rs`?
> 3. Why does the blake3 script have to be split? It seems feasible to run the entire hashing in < 4MB

1. The number of public inputs will affect the cost of the verifier, the more public inputs number, the bigger the MSM computation we need. So generally, we only use the hash as the public input of the proof, and then we can keep constant MSM computation.  Source code in R0 (https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/receipt.rs#L163)
   ```
  /// Verify that this receipt proves a successful execution of the zkVM from the given
    /// `image_id`.
    ///
    /// Uses the zero-knowledge proof system to verify the seal, and decodes the proven
    /// [ReceiptClaim]. This method additionally ensures that the guest exited with a successful
    /// status code (i.e. `Halted(0)`), the image ID is as expected, and the journal has not been
    /// tampered with.
    pub fn verify_with_context(
        &self,
        ctx: &VerifierContext,
        image_id: impl Into<Digest>,
    ) -> Result<(), VerificationError> {
        if self.inner.verifier_parameters() != self.metadata.verifier_parameters {
            return Err(VerificationError::VerifierParametersMismatch {
                expected: self.inner.verifier_parameters(),
                received: self.metadata.verifier_parameters,
            });
        }

        tracing::debug!("Receipt::verify_with_context");
        **self.inner.verify_integrity_with_context(ctx)?;**

        // Check that the claim on the verified receipt matches what was expected. Since we have
        // constrained all field in the ReceiptClaim, we can directly construct the expected digest
        // and do not need to open the claim digest on the inner receipt.
        **let expected_claim = ReceiptClaim::ok(image_id, MaybePruned::Pruned(self.journal.digest()));
        if expected_claim.digest() != self.inner.claim()?.digest() {**
            tracing::debug!(
                "receipt claim does not match expected claim:\nreceipt: {:#?}\nexpected: {:#?}",
                self.inner.claim()?,
                expected_claim
            );
            return Err(VerificationError::ClaimDigestMismatch {
                expected: expected_claim.digest(),
                received: self.claim()?.digest(),
            });
        }

        Ok(())
    }"
```
  "self.inner.verify_integrity_with_context(ctx)?":  Verify the proof with hash;
   "if expected_claim.digest() != self.inner.claim()?.digest()": Verify  hash with public input;
   
   This should be the completed process for ZKP verification. 

 2. we implemented the fragmented version based on the completed version, we want to reserve it to help check the script logic
 
 3. We want to keep the disprove tx smaller to less than 400K in our implementation, that's the reason. 4M is too big to be accepted by the network node other than you get help from them

lucidLuckylee:
Hey sorry I did not come back for this earlier. I think we currently won't use this because all other chunks are not within the same 400k limit. Also this should be the chunkers job

PR #173, opened 2025-01-06
feat: optimize g2 subgroup membership check

### LINKED ISSUE #172: feat: improve G2 subgroup membership check
Current implementation of G2 subgroup membership check splits the total script into 196 chunks. 
With improvements in protocol, denser tapscript and exploiting use of constants, we might be able to reduce the number of such chunks considerably.

manishbista28:
Fixes #172 

Number of chunks reduced from 196 to 26 due to improvements in
- Algorithm: 64 bit scalar used instead of 128 bits halving the cost
- Denser packing: More double and adds per tapscript

Tests included.
Since g2 membership check hadn't and hasn't been integrated to other modules yet, doesn't affect other features.

Depends upon G2 hinted affine add and double functions. If these functions are susceptible to change, will require adapting to those changes as well.


manishbista28:
Used improvement implemented here:
https://github.com/Consensys/gnark-crypto/blob/11d2ed41a3d3470f9cf21a8e1cd4ca5d60a3983f/ecc/bn254/g2.go#L491

```
hinted_msm 0: 3295982 @ 576 stack
hinted_msm 1: 3295982 @ 936 stack
hinted_msm 2: 3295982 @ 936 stack
hinted_msm 3: 3160036 @ 918 stack
hinted_msm 4: 3295982 @ 936 stack
hinted_msm 5: 3295982 @ 936 stack
hinted_msm 6: 3295982 @ 918 stack
hinted_msm 7: 3295982 @ 918 stack
hinted_msm 8: 3160036 @ 900 stack
hinted_msm 9: 3295982 @ 936 stack
hinted_msm 10: 3160036 @ 900 stack
hinted_msm 11: 1715974 @ 702 stack
hinted_msm 12: 1715974 @ 702 stack
hinted_msm 13: 3295982 @ 918 stack
hinted_msm 14: 3160036 @ 900 stack
hinted_msm 15: 3295982 @ 936 stack
hinted_msm 16: 3295982 @ 936 stack
hinted_msm 17: 3295982 @ 918 stack
hinted_msm 18: 3295982 @ 918 stack
hinted_msm 19: 3295982 @ 936 stack
hinted_msm 20: 3295982 @ 936 stack
hinted_msm 21: 3160036 @ 900 stack
hinted_msm 22: 3160036 @ 900 stack
hinted_msm 23: 3295982 @ 936 stack
hinted_p_power_endomorphism 24: 3449600 @ 828 stack 
hinted verify 25: 1443702 @ 540 stack
test bn254::g2_subgroup_check::test::test_is_in_g2_subgroup ... ok
```

Cyimon:
> 3295982

Does this value imply the chunk size?

manishbista28:
> Does this value imply the chunk size?

Yes it is the chunk size.
What it doesn't include is the cost of hashing and bit-commitments.
Since the input and output values to hash are G2Affine, you will have to hash G2Affine twice, which will cost around 350K more.
The cost of bit-commitments will be much low and as such the total size of tapscript will still be less than 4M.

Cyimon:
> > Does this value imply the chunk size?
> 
> Yes it is the script size. What it doesn't include is the cost of hashing and bit-commitments. Since the input and output values to hash are G2Affine, you will have to hash G2Affine twice, which will cost around 350K more. The cost of bit-commitments will be much low and as such the total size of tapscript will still be less than 4M.

Great, thanks for your clarification. Btw, do you think "less than 4M" is still big for the production environment?

manishbista28:
That might indeed be the case. I assumed that non-standard transactions up to 4M are acceptable. However, to ensure the algorithm remains applicable in scenarios where this assumption does not hold, I introduced a “window” parameter. By default, this parameter is set to 4, which means a single tapscript can accommodate approximately G2Affine point operations (both doubling and addition). Reducing this value decreases the size of the tapscript but comes at the cost of requiring a greater number of bitcomms.

The "window" param doesn't apply to endomorphism(second last) chunk though.

manishbista28:
Disclaimer: a [TODO](https://github.com/BitVM/BitVM/blob/09939da10a2ddd85b5d8cb370c4aed8089716427/bitvm/src/bn254/g2_sg_check.rs#L125) here is marked for a case when t == +/- q. Such a case wouldn't be possible in normal execution of Miller loop where t is the G2Affine accumulator. 
But for G2 Subgroup check, i am uncertain at this point if such a case can occur.

How to handle is known:
If t == q, hinted_add is replaced by hinted_double 
and if t == -q, result is G2Affine::zero() without calculation

As far as I know, this problem is not unique to this implementation of subgroup check as the problem stems from point addition operation which is common across subgroup check implementations.

manishbista28:
requesting @lucidLuckylee  @Hakkush-07  to review.

bitvm/src/bn254/g2_sg_check.rs
```diff
@@ -0,0 +1,704 @@
...
+    let mut script = script!();
+    for bit in bits {
+        if bit == 0 {
+            let (scr, hint) = hinted_check_double(acc);
+            hints.extend_from_slice(&hint);
+            script = script!(
+                {script}
+                {Fq2::toaltstack()} {Fq2::toaltstack()} // move q to altstack
+                {scr}
+                {Fq2::fromaltstack()} {Fq2::fromaltstack()} // bring q to altstack
```
lucidLuckylee, line 46: `// bring q to altstack` should probably be `//move q to stack`

bitvm/src/bn254/g2_sg_check.rs
```diff
@@ -0,0 +1,704 @@
...
+            
+        } else if bit == 1 {
+            let (scr, hint) = hinted_check_add(acc, q);
+            hints.extend_from_slice(&hint);
+            script = script!(
+                {script}
+                {Fq2::copy(2)} {Fq2::copy(2)}
+                {Fq2::toaltstack()} {Fq2::toaltstack()} // move q to altstack
+                {scr}
+                {Fq2::fromaltstack()} {Fq2::fromaltstack()} // bring q to altstack
```
lucidLuckylee, line 58: Same here.

lucidLuckylee (approved):

bitvm/src/bn254/g2_sg_check.rs
manishbista28: added understandable comment showing stack snapshot in latest commit.

bitvm/src/bn254/g2_sg_check.rs
manishbista28: added understandable comment showing stack snapshot in latest commit.

lucidLuckylee: Thank you, even better!

PR #177, opened 2025-01-09
Refactor - BN254 G1 and G2 curve operations

Hakkush-07:
This PR splits G1Affine, G1Projective, G2 related functions in curves.rs, utils.rs, g2_subgroup_check.rs into their own g1.rs, g1p.rs, g2.rs files. Also, moves push functions such as g1p_push from utils to their own structs. Closes #171

PR #184, opened 2025-01-10
Refactor - Remove montgomery functions

Hakkush-07:
This PR removes montgomery functions and deletes the "_not_montgomery" suffix from those that are not montgomery. As the effect of it, some unhinted functions, starting with Fq.mul, Fq.square, Fq.inv, are removed. Also, removes u29x9 and u32x8 as they are no longer used. Also removes G1Projective.

Note: It is better to merge this after #177 since this is on top of it.


manishbista28:
will review after merge conflicts have been fixed :) thanks

Hakkush-07:
#188 replaces this

PR #185, opened 2025-01-10
Bridge use global secp256k1 context

eyusufatik:
Creating Secp256k1 context everytime causes precomputed values to be calculated for each signature verification, causing big memory allocations and unnecessary elliptic curve arithmetic.

Used global context provided by the secp256k1 library instead.

Verifier, operator, base context already had their own secp instances which mitigates this problem, however, by using the global instance we simplify and unify the code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ ark-crypto-primitives = { git = "https://github.com/arkworks-rs/crypto-primitive
 ark-relations = { git = "https://github.com/arkworks-rs/snark/" }
 serial_test = "*"
 tqdm = "0.7"
+secp256k1 = { version = "0.29.1", features = ["global-context"]}
 
 [profile.dev]
 opt-level = 3
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1748,6 +1748,7 @@ dependencies = [
  "openssh-sftp-client",
  "rand",
  "regex",
+ "secp256k1",
  "serde",
  "serde_json",
  "serial_test",
--- a/bridge/Cargo.toml
+++ b/bridge/Cargo.toml
@@ -38,7 +38,7 @@ ark-bn254.workspace = true
 ark-groth16.workspace = true
 ark-ff.workspace = true
 ark-std.workspace = true
-
+secp256k1.workspace = true
 
 [profile.dev]
 opt-level = 3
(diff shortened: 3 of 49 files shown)


aoengin (approved):

lucidLuckylee:
Can we use the global-context feature of the [secp256k1 crate](https://docs.rs/secp256k1/latest/secp256k1/)? It provides an API with which we wouldn't have to provide the context to every function.

eyusufatik:
> Can we use the global-context feature of the [secp256k1 crate](https://docs.rs/secp256k1/latest/secp256k1/)? It provides an API with which we wouldn't have to provide the context to every function.

that's the feature I activated. I wasn't aware it also gives you a completely different API. I'll see what we can do 👍

eyusufatik:
@lucidLuckylee pushed the requested changes.

These don't have `global-context` alternatives because they are coming from the bitcoin crate:
- `TaprootBuilder::finalize`
- `PublicKey::from_private_key`


And the `sign_schnorr_no_aux_rand` function has a `global-context` alternative on newer versions of the crate, see my comment.

Cargo.toml
```diff
@@ -50,6 +50,7 @@ ark-crypto-primitives = { git = "https://github.com/arkworks-rs/crypto-primitive
 ark-relations = { git = "https://github.com/arkworks-rs/snark/" }
 serial_test = "*"
 tqdm = "0.7"
+secp256k1 = { version = "0.29.1", features = ["global-context"]}
```
lucidLuckylee, line 53: WIth this and the rust bitcoin reexport we have two dependencies to `secp256k1`. If possible can you patch the rust-bitcoin `secp256k1` reexport instead?

lucidLuckylee:
Since there are functions with the old API we need our own global `SECP` context anyway so we might aswell stick with the original change and not use the `global-context` feature flag. I think it is not worth to bring in the extra dependency for such little benefit. Sorry!

lucidLuckylee:
Why did you need the `global-context` flag in the first place before using the different API?
I think the extra dependency to the secp crate and even patching is should be avoided if we can.

eyusufatik:
> Why did you need the `global-context` flag in the first place before using the different API?

The `global-context` feature exports the static instance `SECP256K1`. If we did not enable this, we could use something like `lazy_static` and create our own "global instance". But with the feature enabled, we can import it and pass it to functions that still need a Secp256k1 instances.

The "new API" is just a few extra functions that are wrappers over the originals, they only pass the static instance `SECP256K1` to the old API.

> I think the extra dependency to the secp crate and even patching it should be avoided if possible

I'm looking at how to patch something that is *only* a subdependency but it might not be doable. Why don't you want the secp256k1 dependency by the way?

lucidLuckylee:
Unfortunately there seems to be no way around the extra dependency (https://users.rust-lang.org/t/adding-features-to-a-dependencys-dependency/106651). I'll merge what you got :) Thank you for the PR!

PR #210, opened 2025-02-04
Constant Pairing

### LINKED ISSUE #202: Replace Constant Pairing in Groth Verifier
Currently the Groth16 verifier code performs four pairings, but one of these pairings is between elements of the verification key. The result of this pairing is constant in the proof so we can just precompute it and save the pairing. Specifically the third pairing between `vk.alpha_g1` and `-vk.beta_g2` can be replaced by multiplication with a precomputed, constant Fp12 element. Seems like this should be a relatively simple change with relatively large savings (three instead of four pairings). This is a standard Groth optimization.

Hakkush-07:
This PR removes calculations for constant pairing and instead checks equality at the end. Closes #202

### DIFF
--- a/bitvm/src/bn254/pairing.rs
+++ b/bitvm/src/bn254/pairing.rs
@@ -70,6 +70,7 @@ impl Pairing {
             }
 
             for (j, p) in p_lst.iter().enumerate().take(num_line_groups) {
+                if j == 2 { continue; }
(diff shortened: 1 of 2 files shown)


Hakkush-07:
This amends the unchunked part, #215 applies a similar fix to chunked part.

Issue #174, opened 2025-01-06
doc: api specs for integration between groth16 verifier and bridge


manishbista28:
Write API spec for integration between bridge and chunked-groth16-verifier.

lucidLuckylee:
I like this spec.

The bridge assumptions allows verifiers to come in just for step 7 without being part of the setup ceremony.
I thought about that and late verifiers could fetch the PublicKeys from the assert transactions' scripts instead of relying on the Bridge Operator or the federation to get the public keys from step 2.
However, we can of course not reconstruct the pre-signed disprove transactions so someone has to distribute them and make them available for verifiers that were not in the setup.

manishbista28:
Agree with using a single public input which is a hash of the actual public inputs. 
One question: what would be the size of this hash-digest ? Asking because if it's 32 bytes and has to be represented as a scalar element (254-bit Fr), some (like risc0) split into two scalars, while some (like succinct) pad the hash with two zeros to limit its range to less than r.

manishbista28:
vk_y0 is only included in only 1 disprove script. [this one](https://github.com/BitVM/BitVM/blob/main/src/chunker/chunk_msm.rs#L77) specifically.

manishbista28:
vk_y0 itself is never bitcommitted or directly used in tapscript. During the compilation step, operators calculate vk_y0' and use it with rest of the verification keys in the rest of the process. So you'd have to run Step 1 for each combination of (vk and hardcoded-scalars) once to obtain vk_new, run compile_partial_disprove_scripts(vk_new) and can reuse the output of this step cheaply.

manishbista28:
> I thought about that and late verifiers could fetch the PublicKeys from the assert transactions' scripts instead of relying on the Bridge Operator or the federation to get the public keys from step 2.
However, we can of course not reconstruct the pre-signed disprove transactions so someone has to distribute them and make them available for verifiers that were not in the setup.

@lucidLuckylee 
yes, late verifiers are able to extract PublicKeys from locking_script (winternitz::checkSigVerify()) committed on Claim Txns and won't have to rely on any distributor. 
Afterwards, late verifiers can also reconstruct disprove_scripts from Step 1 and 2 independently, then run verify_signed_assertions.

lucidLuckylee:
> > I thought about that and late verifiers could fetch the PublicKeys from the assert transactions' scripts instead of relying on the Bridge Operator or the federation to get the public keys from step 2.
[...]

PR #215, opened 2025-02-11
Feat/213 chunked verifier with folded fp12

### LINKED ISSUE #213: Represent fq12 in folded form
given an fq12 element: f = a + b J
Represent f as f/a. You can divide by any Fq6(a) because it cancels out with final exponentiation. This is the same property that we had already benefitted from while representing sparse evaluation, where we divide line evaluation by y_p.

Image

 As such f = 1 + c j. Similarly for g = 1 + dj, we can compute fp12Xfp12 mul as follows
f.g = (1 + cj) * (1 + dj) =   (1+cd. j^2) + (c+d).j = 1 + [(c+d)/(1+cd. j^2)] = 1 + ej
where e = (c+d)/(1+cd. j^2)

Given e, c, and d, verify e. (1 + cd j^2) == (c+d).
This verification requires two Fq6 multiplications compared to 3 mults with Karatsuba like Fq12XFq12 previously.

In Script, along with c and d, an executor will also need to pass e as hint.

manishbista28:
Closes #213


manishbista28:
Need to remove cargo fix on out of scope files like bridge/tests, chunker/ and bn254/std.rs

manishbista28:
The tests inside g16.rs depend need a sequential order of execution as they are written to resemble the steps starting from Compilation to Disprove.
At the moment, unsure how to do it as cargo test run during CI may run it in parallel.

manishbista28:
Major works left for complete end-to-end execution through bridge are
1. Integrate verifier with bridge 
  So far the exposed methods are based on [this discussion](https://github.com/BitVM/BitVM/issues/174). Will integrate it to existing bridge module.
2. Add validity checks to few tapscripts [mentioned here ](https://github.com/BitVM/BitVM/issues/213#issuecomment-2647620336). Is not difficult.
3. Swap out wots.rs with the one already present winterntiz.rs
  wots.rs was sth we had been working on top of. Will look to replace with existing winternitz.

manishbista28:
Finished points 2 and 3 mentioned above.
will like to write a lot of tests for 2, but a bit lesser priority for now.
will work on integration next.

manishbista28:
Screenshot 2025-02-16 at 20 22 51

Test failed because it used too much stack memory. I had got past it by appending RUST_MIN_STACK=104857600 to cargo test. Not sure if it's the right approach here.

manishbista28:
there are some trivial changes made by cargo clippy fix that has increased the number of file changes, would probably be better to revert to reduce clutter.

bitvm/src/chunk/api_compiletime_utils.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub(crate) p3vk: Vec<ark_bn254::G1Affine>,
+    pub(crate) p1q1: ark_bn254::Fq12,
+    pub(crate) vky0: ark_bn254::G1Affine,
+}
+
+pub(crate) fn generate_partial_script(
+    vk: &ark_groth16::VerifyingKey<Bn254>,
+) -> Vec<bitcoin_script::Script>  {
+    println!("generate_partial_script");
+    assert!(vk.gamma_abc_g1.len() == NUM_PUBS + 1); // supports only 3 pubs
```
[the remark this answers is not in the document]
manishbista28: yes the comment "//supports only 3 pubs" is outdated.
It can support larger number of NUM_PUBS but that would mean the number of tapscripts required for MSM would also increase, albeit by a small amount. There is not restriction on how many such public inputs can be supported.

bitvm/src/signatures/wots_api.rs
```diff
@@ -0,0 +1,150 @@
...
+                }
+
+                /// Create a verification script for a WOTS public key.
+                pub fn checksig_verify(public_key: PublicKey) -> Script {
+                    let params = winternitz::Parameters::new(MSG_LEN * 2, 4);
+                    WINTERNITZ_MESSAGE_VERIFIER.checksig_verify(&params, &public_key.to_vec())
+                }
+
+                /// Generate a signature for a message using the provided secret.
+                pub fn get_signature(secret: &str, msg_bytes: &[u8]) -> Signature {
```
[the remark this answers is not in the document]
manishbista28: WINTERNITZ_MESSAGE_VERIFIER is a ListpickVerifier whose signature is basically HASH160 and message interleaved. It meets the requirement of Assert.
For disprove, we remove the message part and only keep the HASH160 part (also called compact form). This is then used in Disprove Script with WINTERNITZ_MESSAGE_COMPACT_VERIFIER.

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+            {Fq::fromaltstack()}
+            {Fq::equal(1, 0)}
+            if msg_index == elem_types.len()-1 {
+                // [should_do_output_validity_check_bit, output_is_equal]
+                OP_TOALTSTACK // [should_do_output_validity_check_bit] [output_is_equal]
+                OP_IF
+                    OP_FROMALTSTACK
+                    OP_NOT  // disprove if claimed output and hashed output are not equal
+                OP_ELSE
```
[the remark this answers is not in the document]
manishbista28: the basic intuition is if the input that was bitcomitted isn't valid (e.g. say not a field element), then you can directly meet Disprove condition, you do not need to compare output hashes. Sometimes it may not be possible to even generate output to be able to compare (e.g. you can't compute say inverse of a field element if the input is not a field element to begin with).

To allow for this condition, we include at the top of the stack a bit 0/1 for should_do_output_validity_check_bit.
I have also added a bit more information on README.md's Disprove Logic Section.

We can not put this condition in rust code because the value of this bit is known only after execution (i.e. after checking validity of bitcomitted input, thus is not known in advance).

bitvm/src/chunk/elements.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn test_hash_witness() {
+        let mut prng = ChaCha20Rng::seed_from_u64(117);
+        let fld = ark_bn254::Fq6::rand(&mut prng);
+        let elem = super::DataType::Fp6Data(fld);
+
+        let preim = elem.to_witness(ElementType::Fp6);
+        let scr = script!(
+            for p in preim {
+                {p.push()}
+            }
```
manishbista28, line 339:
```
    #[test]
    fn test_hash_witness() {
        let mut prng = ChaCha20Rng::seed_from_u64(117);
        let fld = ark_bn254::Fq6::rand(&mut prng);
        let elem = super::DataType::Fp6Data(fld);

        let preim = elem.to_witness(ElementType::Fp6);
        let check_output_bit = 1;
        let scr = script!(
            for p in preim {
                {p.push()}
            }
            {elem.to_hash().as_hint_type().push()}
            {Fq::toaltstack()}
            {check_output_bit}
            {hash_messages(vec![ElementType::Fp6])}
        );
        let res = execute_script(scr);
        assert!(!res.success && res.final_stack.len() == 1);
    }
    
```
    check_output_bit was missing from this test after recent update; will add.

manishbista28:
TODOs 
(Will add more as review progresses and fix them along. Can move less urgent ones for another PR even)

- move g16.rs tests to api.rs, make it functional and execute serially
- remove panics, add Result<>
- add more test to wrap_hasher.rs, fix the one that's broken (commented above)
- api/generate_signatures() should take array of secrets

AaronFeickert:
It may be worthwhile to set up a test group for slow-running tests. These could be excluded from CI and instead run on a cron.

bitvm/src/signatures/wots_api.rs
```diff
@@ -0,0 +1,150 @@
...
+                        let sigs = WINTERNITZ_MESSAGE_COMPACT_VERIFIER.sign(
+                            &winternitz::Parameters::new(MSG_LEN * 2, 4),
+                            &secret_key,
+                            &msg_bytes.to_vec(),
+                        );
+                        assert_eq!(sigs.len(), N_DIGITS as usize);
+
+                        let mut sigs_vec: Vec<[u8; 20]> = Vec::new();
+                        // Iterate over the signature pieces using step_by.
+                        for i in (0..sigs.len()).step_by(2) {
```
manishbista28, line 132: should be step_by(1) because compact signature doesn't have interleaving messages

manishbista28: wots_compact::get_signature() isn't used by any other module. so this bug doesn't affect existing function.

bitvm/src/chunk/README.md
```diff
@@ -0,0 +1,261 @@
...
+g16_runner_utils - set of functions that wrap tapscripts and is required by g16_runner_core
+
+primitives - utility functions [will likely refactor]
+taps_msm - tapscripts for MSM
+taps_mul - tapscripts for Fp12 multiplications
+taps_point_ops - tapscripts for point operations
+taps_premiller - miscellaneous tapscripts external to Miller Loop
+
+wrap_hasher - wrapper to blake3_u4 hasher
+wrap_wots - wrapper to winternitz methods
```
storopoli, line 35:
```suggestion
- `api` - interface between verifier and external module (like bridge)
- `api_compiletime_utils` - interfacing methods used during bitvm compile time i.e. presigning
- `api_runtime_utils` - interfacing methods used during bitvm runtime i.e. assertion and disprove
- `assigner` - interface to assign Assertions [to support legacy code]
- `elements` - data structure to represent inputs and outputs of each chunk

- `g16_runner_core` - function to generate execute chunked groth16 verifier and collect necessary information
- `g16_runner_utils` - set of functions that wrap tapscripts and is required by `g16_runner_core`

- `primitives` - utility functions [will likely refactor]
- `taps_msm` - tapscripts for MSM
- `taps_mul` - tapscripts for Fp12 multiplications
- `taps_point_ops` - tapscripts for point operations
- `taps_premiller` - miscellaneous tapscripts external to Miller Loop

- `wrap_hasher` - wrapper to `blake3_u4` hasher
- `wrap_wots` - wrapper to `winternitz` methods
```

Needs to be a list and with proper code formatting in `backticks`

bitvm/src/groth16/g16.rs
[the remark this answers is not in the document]
lucidLuckylee: The tests should be independent. 
You can put them all together in a single test to execute them in order or if you want multiple tests they should each reproduce the required steps from the other tests.

bitvm/src/signatures/wots_api.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Trait for converting a signature into a Script.
+pub trait SignatureImpl {
+    fn to_script(self) -> Script;
+    fn to_compact_script(self) -> Script;
+}
+
+/// Macro that implements the WOTS module for a given message length (in bytes).
+/// For example:
+/// - For 160-bit WOTS, use 20 bytes.
+/// - For 256-bit WOTS, use 32 bytes.
```
lucidLuckylee, line 17: This isn't true yet, right? Currently everything seems hardcoded to 20 bytes.

bitvm/src/bn254/g1.rs
```diff
@@ -305,9 +306,9 @@ impl G1Affine {
 
     
     fn hinted_scalar_decomposition(k: ark_bn254::Fr) -> (Script, Vec<Hint>) {
-        const LAMBDA: ark_bn254::Fr = MontFp!("21888242871839275217838484774961031246154997185409878258781734729429964517155");
+        let lambda: ark_bn254::Fr = ark_bn254::Fr::from(BigUint::from_str("21888242871839275217838484774961031246154997185409878258781734729429964517155").unwrap());
```
lucidLuckylee, line 309: `Lambda` should stay a const.

lucidLuckylee (commented):
Do you want this merged now? I'd rather merge this only after it has been integrated into the current chunker (or entirely replaces it).

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Given an array of Elements of different ElementType, the function assumes the order [Input_A_Preimage, Input_B_Preimage, Output_Preimage] for main stack
+/// and [Output_Hash, Input_B_Hash, Input_A_Hash] for altstack
+// should_do_output_validity_check_bit: we have a bit at the top of stack that determines whether the final equality check between calculated and claimed output
+// should be done. This is used in the context where tapscript is disproven based on input invalidity indifferent to output.
+// i.e. if the claimed input is invalid, user can disprove indifferent to output computation
+pub fn hash_messages(elem_types: Vec<ElementType>) -> Script {
+    // Altstack: [Hc, Hb, Ha]
+    // Stack: [a, b, c, should_do_output_validity_check_bit]
+    let elem_types: Vec<ElementType> = elem_types.into_iter().filter(|et| et.number_of_limbs_of_hashing_preimage() > 0).collect();
+    let mut loop_script = script!();
```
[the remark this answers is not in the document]
manishbista28: will update to this convention

bitvm/src/signatures/wots_api.rs
manishbista28: yeah, we have a macro to generate wots for arbitrary number of bytes, but the chunk/ source code has hardcoded effect of 20 bytes hash in most places. I will switch over to a variable to make it cleaner and valuable when it's required to switch over to lower hash lengths.

manishbista28:
CI is working now and most of the review requests have been addressed.

bitvm/src/groth16/g16.rs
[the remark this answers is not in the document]
manishbista28: moved the tests to chunk/api.
The tests that have serial dependency are marked with #[ignore] flag to skip them from CI.
There exists a single tests called full_e2e_execution() that integrates all of these ignored tests and runs them in CI.
That way all of the intended functionality is run by CI and granular tests written to help with development testing also stays.

bitvm/src/chunk/api_compiletime_utils.rs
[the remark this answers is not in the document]
manishbista28: will cleanup old comments like this in the next commit.

bitvm/src/chunk/elements.rs
manishbista28: added the fix in latest commit

bitvm/src/chunk/README.md
manishbista28: thanks will include it. I was also thinking of moving the entire docs to a separate folder because having this README here seems somewhat out of place and there will be a lot of docs soon to be added.

bitvm/src/chunk/wrap_hasher.rs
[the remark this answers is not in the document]
manishbista28: updated in the latest commit

bitvm/src/bn254/g1.rs
manishbista28: if it's okay, i'd like to defer this to a cleanup PR after this one.
- I get a rust-analyzer error: "proc macro panicked: could not parse" whenever I use MontFp macro. Build and runs works just fine, only a problem with rust-analyzer. 
- Secondly assigning const to value returned by Fr::from() doesn't seem possible at the moment.
```
cannot call non-const fn `<ark_ff::Fp<MontBackend<FrConfig, 4>, 4> as From<BigUint>>::from` in constants
calls in constants are limited to constant functions, tuple structs and tuple variants
```
There are few other instances too where const should be used. Will do them in batch in cleanup PR if okay.

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -0,0 +1,127 @@
...
+    let mut stack = StackTracker::new();
+    blake3_u4_compact(&mut stack, 192, true, true);
+    wrap_scr(stack.get_script())
+}
+
+pub fn hash_448b() -> Script {
+    let mut stack = StackTracker::new();
+    blake3_u4_compact(&mut stack, 448, true, true);
+    wrap_scr(stack.get_script())
+}
```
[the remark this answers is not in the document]
manishbista28: added fix for this in the latest commit

bitvm/src/chunk/README.md
manishbista28: added fix for this in the latest batch commit.

bitvm/src/signatures/wots_api.rs
manishbista28: added fix for this

bitvm/src/signatures/wots_api.rs
manishbista28: cleaned up as many places inside chunk/ to make it refer to a global hash size variable, but that will not be enough to easily make the switch to a different hash size. Being able to switch to a different sized hash is sth doable and valuable. will want to defer that work in a separate PR.

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// This function is used to add hashing layer to the disprove script.
+/// The additional logic here is due to a constraint in blake3_u4 
+/// which requires the entire stack to only hold the message to be hashed.
```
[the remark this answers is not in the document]
manishbista28: This constraint about requiring the stack to only contain message to hash was introduced in PR #136 as part of an optimization.
I believe there isn't an easy enough parameter to disable this constraint. 

Its effect has been bounded to within this function. 

Should things change in future (i.e. not require stack to only contain msg-to-hash), this method will still be applicable, though slightly inefficient (from the cost of moving between stacks).

bitvm/src/groth16/g16.rs
```diff
@@ -0,0 +1,571 @@
...
+            let mut scr: HashMap<u32, Vec<Script>> = HashMap::new();
+            let f = read_map_from_file(file).unwrap();
+            for (k, v) in f {
+                let vs: Vec<Script> = v
+                    .into_iter()
+                    .map(|x| {
+                        let sc = script! {};
+                        let bf = ScriptBuf::from_bytes(x);
+                        
+                        sc.push_script(bf)
```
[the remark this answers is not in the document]
lucidLuckylee: @manishbista28 I added serde serialization to our `Script` type so you can do
```
let binary_data = bincode::serialize(&script).unwrap();
let deserialized: StructuredScript = bincode::deserialize(&binary_data).unwrap();
```
It is quite inefficient but I am thinking about reworking the data structure to use up less memory. So after that is done it will require less space.

PR #217, opened 2025-02-14
Refactor Signatures and Hash

just-erray:
This PR
- Removes unnecessary code
- Adds comments explaining the function of each function
- Shortens the viable parts
- Cleans up and refactors code for clarity
- Adds and edits some functions for future use 

in signatures folder and the blake3_u4.rs file. 
(It doesn't cause any incompatibilities with the [active work](https://github.com/BitVM/BitVM/pull/215).)

### DIFF
--- a/bitvm/src/hash/blake3_u4.rs
+++ b/bitvm/src/hash/blake3_u4.rs
@@ -1,18 +1,21 @@
 use std::collections::HashMap;
-
 use bitcoin_script_stack::stack::{StackTracker, StackVariable};
-
-pub use bitcoin_script::script;
 pub use bitcoin_script::builder::StructuredScript as Script;
-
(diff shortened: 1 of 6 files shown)


lucidLuckylee:
Hi @just-erray, is the current plan to use the blake3_u4.rs primitices in blake3.rs eventually to switch to the u4 version? Wondering, because this PR removes the hashing from blake3_u4.rs but doesn't integrate its primitives into blake3.rs yet.

just-erray:
Hello, I believe the only BLAKE3 version we plan to use is the u4 version with the compact representation (which is supposed to be the case with the new chunker work, though @manishbista28 knows more about it). Since the compact version's hashing function serves the same purpose as the blake3_u4.rs's hashing function, I thought that it was redundant and removed it. I think the current plan is to remove all BLAKE3-related code except for blake3_u4_compact.rs and blake3_u4.rs, as the compact version depends on its functions.

PR #231, opened 2025-02-26
Optimize small-constant uses of OP_ROLL in the bn254 module.

aweinstock314:
Optimizing `OP_ROLL` usage for fq/fr in the `bn254` implementation decreases the total Groth16 verifier size from 931617108 to 929649780 (as reported by `cargo test --release -- groth16::test::test_hinted_groth16_verifier --nocapture --exact` ), a 0.2% improvement.


AaronFeickert (approved):
LGTM, but it was brought up elsewhere that this could perhaps be baked into the compiled more directly.

lucidLuckylee:
Yes I think it is best not to merge this here and to add this to the compiler to make sure the optimization is consistently applied to all scripts.

PR #236, opened 2025-02-28
Refactor squaring test for clarity

AaronFeickert:
Recent work in #215 removes a redundant check involved in folded `Fq12` squaring operations. It modifies a corresponding test such that a previously-invalid zero condition is now considered valid, but retains structure and naming that does not reflect the change.

This PR refactors accordingly for clarity.


manishbista28:
looks good to me, thanks for the correction

PR #239, opened 2025-03-01
test: improve test coverage for sha256

### LINKED ISSUE #212: Improve test coverage for SHA256: Include tests against standard test vectors and random inputs.
All of the current three implementations of SHA256 are not being tested against the [standard test vectors](https://www.di-mgt.com.au/sha_testvectors.html). The tests against the reference implementation for random strings could also be added. 

The test harness for doing the above mentioned tests is already present in [sha256_u4.rs](https://github.com/BitVM/BitVM/blob/89558b3a50e7d08a5eb42aa7de266e7912df3f98/bitvm/src/hash/sha256_u4.rs#L505) and [sha256_u4_stack.rs](https://github.com/BitVM/BitVM/blob/89558b3a50e7d08a5eb42aa7de266e7912df3f98/bitvm/src/hash/sha256_u4_stack.rs#L786). This will make adding the tests a lot easier.

Fuzz testing can be added as well. Targets can be created similar to BLAKE3. The only difference being that conversion of bytearray produced by Fuzzer to hex_array would be needed.

piotr-roslaniec:
Hello maintainers of BitVM. I was just trying my hand at contributing to this repository. Let me know how I can be helpful. I look forward to your review.
- Fixes #212 
- I could not run the fuzzer locally due to `rust-lld: error: undefined symbol: __sancov_gen_.xxxx` errors
- One of the official vectors is hashed twice because it's too long to be used in tests (time and memory overhead). We can still run it if we lift `assert!(num_bytes < 512);` assertion from `sha256.rs` implementation


mukeshdroid:
great work @piotr-roslaniec. Thank you for the PR.

I ran the fuzz test you wrote locally and it does compile. I would love to help you with the getting the fuzz tester running if you can provide more logs. The fuzz tests does fail for now due to input size being greater than 512 bytes.
image

I will try again with the lengths of the inputs restricted. Thank you once again.

mukeshdroid:
I spun up a fresh linux machine and am getting the same errors despite multiple attempts. A minimal example program as provided in the [cargo fuzz tutorial ](https://rust-fuzz.github.io/book/cargo-fuzz/tutorial.html) does compile and work as intended.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,60 @@
...
+        serde_json::from_reader(reader)?;
+    Ok(vectors)
+}
+
+/// Prepare test vector for use in testing. If the input is longer than 512 bytes, it will be hashed twice.
+pub fn prepare_test_vector(data: &str, count: u64, expected_hex: &str) -> (String, String) {
+    let mut input = data.repeat(count as usize).as_bytes().to_vec();
+    let mut expected = hex::decode(expected_hex).unwrap();
+
+    // Some of the test vectors are longer than 512 bytes, so we need to hash them twice
```
AaronFeickert, line 37: I get the idea here, but it's no longer accurate to call these test vectors in the same manner as the others. I would just drop them.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,60 @@
...
+        input = reference_sha256(&input);
+        expected = reference_sha256(&expected);
+    }
+
+    (hex::encode(&input), hex::encode(&expected))
+}
+
+/// Generate random test cases for SHA256
+pub fn random_test_cases() -> Vec<(String, String)> {
+    let test_lengths = [1, 8, 16, 32, 55, 56, 57, 63, 64, 65, 127, 128, 129];
```
AaronFeickert, line 49: Would recommend going up to 512 bytes if feasible.

fuzz/fuzz_targets/sha256.rs
```diff
@@ -0,0 +1,33 @@
...
+impl<'a> Arbitrary<'a> for LimitedBytes {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        let size = u.int_in_range(0..=512)?;
+        let mut bytes = vec![0u8; size];
+        u.fill_buffer(&mut bytes)?;
+        Ok(LimitedBytes(bytes.to_vec()))
+    }
+}
+
+fuzz_target!(|data: LimitedBytes| {
```
AaronFeickert, line 22: I'm getting failed tests on this.

AaronFeickert (changes requested):

AaronFeickert:
Recommend adding randomized testing to `sha256_u4_stack` to bring it in line with the other implementations.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,60 @@
...
+    Ok(vectors)
+}
+
+/// Prepare test vector for use in testing. If the input is longer than 512 bytes, it will be hashed twice.
+pub fn prepare_test_vector(data: &str, count: u64, expected_hex: &str) -> (String, String) {
+    let mut input = data.repeat(count as usize).as_bytes().to_vec();
+    let mut expected = hex::decode(expected_hex).unwrap();
+
+    // Some of the test vectors are longer than 512 bytes, so we need to hash them twice
+    let double_hashing_needed = input.len() > 512;
```
AaronFeickert, line 38: There's an off-by-one error here. The implementation will panic if the input length is 512 bytes.

fuzz/fuzz_targets/sha256.rs
```diff
@@ -0,0 +1,33 @@
...
+use bitvm::hash::sha256::{reference_sha256, test_sha256_with};
+use bitvm::hash::sha256_u4::test_sha256_u4_with;
+use bitvm::hash::sha256_u4_stack::test_sha256_u4_stack_with;
+
+#[derive(Debug)]
+struct LimitedBytes(Vec<u8>);
+
+impl<'a> Arbitrary<'a> for LimitedBytes {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        let size = u.int_in_range(0..=512)?;
```
AaronFeickert, line 15: There is an off-by-one error here. The implementation will panic if the input length is 512 bytes.

bitvm/src/hash/sha256_test_utils.rs
piotr-roslaniec: Agreed. To clarify, there's just one vector that doesn't fit.

AaronFeickert: In that case, I'd say just add a check on the test vector sizes. Randomized testing would be likely to catch the same sorts of errors that might be caught by this double-hashing method anyway, and with more clarity of intent.

bitvm/src/hash/sha256_test_utils.rs
AaronFeickert: If this ends up slowing down CI significantly, the list could probably be pruned a bit as well.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,76 @@
...
+    (1..=511)
+        .map(|len| {
+            let random_bytes: Vec<u8> = (0..len).map(|_| rand::random::<u8>()).collect();
+            let input_hex = hex::encode(&random_bytes);
+            let expected = reference_sha256(&random_bytes);
+            let expected_hex = hex::encode(expected);
+            (input_hex, expected_hex)
+        })
+        .collect()
+}
```
AaronFeickert, line 76: I get failures for all three SHA256 variants against this.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,76 @@
...
+        127, 128, 129,
+        // 254  // fails
+        // 255, // fails
+        // 256, // fails
+        // 257, // fails
+
+        // Near maximum supported size
+        // 510, // fails
+        // 511, // fails
+    ];
```
AaronFeickert, line 64: These aren't used.

AaronFeickert (changes requested):

piotr-roslaniec:
Hi @AaronFeickert, thank you for looking into this.

After adding longer test vectors, some tests started to fail:
```
sha256 test failed for input: <ommited> with length 368
sha256_u4 test failed for input: <ommited> with length 288
sha256_u4_stack test failed for input: <ommited> with length 398
```
I ran these tests a couple of times, and they fail with different outputs with the same length.

I will investigate this issue.

AaronFeickert:
@mukeshdroid and I looked into it further, and identified that a more restrictive limit of 128 input bytes seems to do the trick, passing both randomized unit testing and fuzzing. The current use case for this functionality doesn't put us over 64 bytes, so the limit seems reasonable to us. This also allows for faster randomized unit testing that hits all applicable lengths.

piotr-roslaniec:
Hi @AaronFeickert. I submitted the updated tests. All tests have reasonable execution times (<20s), except for `hash::sha256_u4_stack::tests::test_sha256_random`, which takes several minutes (~480s) to run.

Thank you for the review and your continued support.

AaronFeickert (commented):
Just a nit, but I'd say to make this limit inclusive for clarity: you should be able to provide 128 bytes of input (not only 127).

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    let mut prepared_vectors = vec![];
+    for vector in vectors.iter() {
+        let input = vector
+            .message
+            .data
+            .repeat(vector.message.count as usize)
+            .as_bytes()
+            .to_vec();
+        if input.len() >= 128 {
```
AaronFeickert, line 36:
```suggestion
        if input.len() > 128 {
```

AaronFeickert: Otherwise the bound is off.

bitvm/src/hash/sha256_u4_stack.rs
```diff
@@ -984,4 +996,31 @@ mod tests {
...
+    #[test]
+    fn test_sha256_random() {
+        for (input_hex, expected_hex) in random_test_cases() {
+            test_sha256_u4_stack_with(&input_hex, &expected_hex, true, true);
+            test_sha256_u4_stack_with(&input_hex, &expected_hex, true, false);
+            test_sha256_u4_stack_with(&input_hex, &expected_hex, false, true);
+            test_sha256_u4_stack_with(&input_hex, &expected_hex, false, false);
+        }
+    }
 }
```
AaronFeickert, line 1026: Recommend splitting these four cases out to separate tests for efficiency.

The current test format takes 272 s on my machine. Splitting it up into four tests (and running across four cores) takes a total of 110 s. Interestingly, the same splitting using `cargo nextest` takes only 80 s.

The [standard runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) has four cores, so this seems reasonable to me.

AaronFeickert (changes requested):
Definitely coming together nicely!

bitvm/src/hash/sha256_u4_stack.rs
mukeshdroid: I get similar test times (~ 2 mins) even when the test cases are split. 

If we wanna make it faster, we could sample random strings of varying length between 1 and 128 rather than testing for strings of all lengths in the range 1 to 128. This is what was done for BLAKE3 since there the input limit was 1024 bytes and hence testing on 1024 random strings was infeasible.

mukeshdroid:
If we are limiting our testing to inputs of size 128 bytes, we should place an assertion that validates that inputs to these functions are indeed smaller than or equal to 128 bytes.

On the other hand, we could determine the input limits for the specific implementations.

For the sha256_u4_stack variant, estimating the maximum supported input length based on the stack limit of 1000 worked. The calculations are in the spreadsheet. I tested these limits using a single random string and the result was as expected. A random string of length 198 was hashed correctly when both parameters of sha256_u4_stack is set to true while it failed on an input of 199 bytes.

For the sha256 and sha256_u4 variants, the same calculations for maximum allowed input size doesn't hold. Testing out with random strings, the maximum input length is 183 bytes and 143 bytes respectively.

Instead of 128 bytes, we could place these values as assertion to check input validity and add tests for these input limits in unit tests as well as fuzz tests. 

This might not be fruitful considering the effort if we plan to move away from sha256 in the future though.

AaronFeickert:
This will likely need to be rebased and formatted, since #252 enforces formatting.

piotr-roslaniec:
Thank you for your patience and guidance in implementing these changes.

I rebased and formatted this PR.

bitvm/src/hash/sha256.rs
```diff
@@ -28,12 +31,18 @@ const INITSTATE: [u32; 8] = [
 
 /// sha256 take indefinite length input on the top of stack and return 256 bit (64 byte)
```
AaronFeickert, line 32: This docstring is no longer accurate, since you cap the input length.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,67 @@
...
+    let mut prepared_vectors = vec![];
+    for vector in vectors.iter() {
+        let input = vector
+            .message
+            .data
+            .repeat(vector.message.count as usize)
+            .as_bytes()
+            .to_vec();
+        if input.len() > input_limit {
+            // The implementation only supports inputs up to 128 bytes
```
AaronFeickert, line 40: This limit is wrong, no? It depends on the implementation, hence it being a function parameter.

bitvm/src/hash/sha256_test_utils.rs
```diff
@@ -0,0 +1,67 @@
...
+        prepared_vectors.push((hex::encode(&input), hex::encode(&expected)));
+    }
+
+    Ok(prepared_vectors)
+}
+
+/// Generate random test cases for SHA256
+pub fn random_test_cases(input_limit: usize) -> Vec<(String, String)> {
+    // We return [1, max_len] and at most 100 test cases in between
+    let num_cases = (input_limit / 4) % 100;
```
AaronFeickert, line 53: Why randomize the number of tests in this way?

bitvm/src/hash/sha256_u4.rs
```diff
@@ -2,6 +2,9 @@ use crate::treepp::{script, Script};
 use crate::u4::{u4_add::*, u4_logic::*, u4_rot::*, u4_std::*};
 use std::vec;
 
+/// A pre-calculated limit on the size of input
+pub const INPUT_N_BYTES_LIMIT: usize = 143;
+
```
AaronFeickert, line 7: This limit doesn't appear to be enforced.

fuzz/fuzz_targets/blake3.rs
AaronFeickert, line 1: I assume this file's changes are just to make the formatter happy?

AaronFeickert (changes requested):
Sorry for the delay on this review. Just some minor nits and questions remaining.

fuzz/fuzz_targets/sha256.rs
AaronFeickert, line 1: Looks like you'll need to refactor this to put each fuzz target in a separate file with separate configuration. Apparently `cargo fuzz` does not support including multiple targets.

AaronFeickert (changes requested):

PR #248, opened 2025-03-07
Refactor BLAKE3 implementations

### LINKED ISSUE #221: Refactor(blake3): Switch to blake3_u4_compact version
We only want to keep
- `blake3.rs` as the main API
- `blake3_u4_compact.rs` which provides the actual u4 blake 3 hashing
- `blake3_u4.rs` which includes all necessary u4_primitives used by `blake3_u4_compact.rs`

Currently blake3_u4 due to the stack tracker requires the stack to be completely empty which is a constraint we would ideally get rid of to ensure the API is stable.

AaronFeickert:
This PR refactors and simplifies the BLAKE3 implementations. Specifically, the goal is to move exclusively to `blake3_u4_compact`.

Closes #221.


AaronFeickert:
Note that this is EXTREMELY BREAKING from a public API perspective.

manishbista28 (approved):

wz14:
It's more readable and doesn't affect my side. LGTM.

just-erray:
We are using the non-compact blake3_u4 version in our bridge design, but it was previously deleted since it isn't used on the BitVM side and was added on top of our branch. This PR makes the implementation of the compact version more organized and readable, and we can edit our branch accordingly, so it works perfectly for us.

mukeshdroid (approved):

uncomputable (approved):
ACK 4342aa66f095b9443857d5cee4c61804438e803f All changes are removals, code moves, renamings, or changes in visibility.

PR #252, opened 2025-03-11
Enable rustfmt in CI

wz14:
Hope to enable rustfmt in CI. 
I think it's okay to remove the `ignore` field in `.rustfmt.toml`. 
Please double-check that.


AaronFeickert:
I like the idea in theory, but am wary in practice due to the complexity of the changes.

This would probably require that multiple reviewers run the formatter locally themselves and do an automated diff to ensure these changes match. Otherwise I wouldn't be comfortable merging.

wz14:
> I like the idea in theory, but am wary in practice due to the complexity of the changes.
> 
> This would probably require that multiple reviewers run the formatter locally themselves and do an automated diff to ensure these changes match. Otherwise I wouldn't be comfortable merging.

Well, I don't think it's necessary to get multiple reviewers to work on this because CI itself is to solve this type of problem. I added `cargo fmt --all -- --check` to the CI workflow. 

Thanks for reminding me that we seem not to specify the rustc version yet. Different version's fmt tools may make different changes. Rust 1.85 is the latest version, how do you think of this version?

AaronFeickert:
Adding formatting to CI will certainly ensure that this PR handles all formatting correctly. The extra checks would be to ensure that nothing else was introduced, which the new CI can't catch and which is hard to inspect manually with this many changes.

AaronFeickert (approved):
Confirmed using an automated diff that formatting is done correctly, and that no other changes (aside from the CI job and `ignore` removal) are present.

AaronFeickert:
@wz14: I'm fine with using stable for now.

AaronFeickert:
This was force pushed, so my previous review is void.

wz14:
Thanks. @AaronFeickert
Reviewers who want to check this PR could run these commands to diff.
```
git checkout 2d6f499  # checkout to the version after changing CI and .rustfmt.toml
cargo fmt
git diff 7e61339
```

AaronFeickert (approved):
LGTM

AaronFeickert:
On second thought, recommend having this run on draft too. Formatting is speedy.

AaronFeickert (approved):
Good call.

PR #255, opened 2025-03-12
feat: support generic hash length in assertion

### LINKED ISSUE #251: Reduce 160-bit hashes to 128 bits
We currently use 160-bit outputs for BLAKE3. Since we only require preimage/second-preimage resistance and not full collision resistance for these hashes, there is nothing to be gained while targeting a 128-bit security level. Reducing to 128-bit outputs means correspondingly smaller Winternitz keys.

It was also discussed that because BN254 is currently thought to provide no more than ~100 bits of security in practice, it may be possible to reduce to 100-bit hashes where a curve attack would be more attractive to an attacker. This seems trickier to justify and more subtle to reason about, however.

manishbista28:
Closes #251 
Support arbitrary hash lengths for chunk outputs.
Default hash length hash been set to 16 bytes.

cargo test -- --skip bridge:: --skip tests::test_final_circuit
bridge::e2e::peg_out::disprove_success::test_e2e_disprove_success
bridge::e2e::peg_out::disprove_reject::test_e2e_disprove_reject

tested above for both 20 and 16 bytes


bitvm/src/signatures/wots_api.rs
```diff
@@ -144,7 +144,9 @@ macro_rules! impl_wots {
...
     };
 }
+const BIGINT_LEN: u32 = 32;
+pub const HASH_LEN: u32 = 20; // bytes, can lower it to value like 16 or 13 lesser acceptable security
 
 // Expand the macro for the two variants.
-impl_wots!(wots160, 20);
-impl_wots!(wots256, 32);
\ No newline at end of file
+impl_wots!(wots_hash, HASH_LEN);
```
manishbista28, line 151: this is the only significant change in this PR.
The rest are naming changes from _160 to _hash.
To change to a lower hash length, only change HASH_LEN param above.
example: set HASH_LEN: u32 = 16 for 16 byte hash.

manishbista28:
what should be the default value ? right now it's 20 bytes.

manishbista28:
the default value has now been set to 16 bytes for 128-bits second preimage resistance.
any implementations can change it to a different value by altering wots_api::HASH_LEN parameter.

bridge/src/commitments.rs
```diff
@@ -109,10 +109,10 @@ impl CommitmentMessageId {
                 WinternitzSecret::new(32),
             );
         }
-        for i in 0..NUM_U160 {
+        for i in 0..NUM_HASH {
             commitment_map.insert(
-                CommitmentMessageId::Groth16IntermediateValues((format!("{}", i + NUM_PUBS + NUM_U256), 20)),
-                WinternitzSecret::new(20),
+                CommitmentMessageId::Groth16IntermediateValues((format!("{}", i + NUM_PUBS + NUM_U256), wots_api::HASH_LEN as usize)),
+                WinternitzSecret::new(wots_api::HASH_LEN as usize),
```
manishbista28, line 115: bridge module references generic hash length variable instead of hardcoding.

AaronFeickert (approved):
LGTM, pending CI.

AaronFeickert:
Agreed that 16 bytes should be used for now. It may be the case that further analysis indicates we can drop down to 13 bytes (as noted in #251), but that's more subtle and deserves careful thought.

just-erray (approved):

AaronFeickert (approved):

AaronFeickert (approved):

manishbista28:
all tests pass

manishbista28:
all tests passed locally.
applied `cargo fix` and `cargo clippy --fix --lib -p bitvm ` which increased the number of line changes unfortunately.

AaronFeickert (approved):

PR #257, opened 2025-03-12
Refactor blake3

### LINKED ISSUE #254: Speed up BLAKE3 fuzzing
It was noted that BLAKE3 fuzzing is hindered by regenerating the script on each input, which is unnecessary. Providing functionality to expose the script and use it directly would provide a significant efficiency boost.

uncomputable:
This PR further cleans up the BLAKE3 code. It adds functions for pushing a message input onto the stack, for computing BLAKE3 on that input, and for verifying that the resulting output matches the expected hash. The unit tests are rewritten using this logic. The fuzz test is rewritten to use a cached version of the BLAKE3 compute script in each iteration.

Fixes #254 

## Need for half tables?

This PR removes the option for half tables from the API. I argue that we should always use full tables (smaller script size, more stack use) and not burden the caller with deciding which table to use. As far as I can see, BLAKE3 [is always called with full tables in BitVM](https://github.com/BitVM/BitVM/blob/0c4d463c3e2dd79d3b27d6958f5ebff6e24b592f/bitvm/src/chunk/wrap_hasher.rs#L23).

~~If there is consensus that we don't need half tables, then I will go ahead and remove that code. I will keep this PR as a draft until we have resolved this question.~~
There seems to no one who needs half tables. With the given BLAKE3 implementation, the stack has to be empty except for the message. In this setting, the script has access to the maximum number of stack items. This nullifies the upside of half tables and makes full tables always better.

### DIFF
--- a/bitvm/src/hash/blake3.rs
+++ b/bitvm/src/hash/blake3.rs
@@ -1,16 +1,13 @@
 use std::collections::HashMap;
 
-use bitcoin::hex::DisplayHex;
-use bitcoin::ScriptBuf;
+use bitcoin::hex::FromHex;
 use bitcoin_script_stack::stack::StackTracker;
-use hex::FromHex;
 use itertools::Itertools;
 
 pub use bitcoin_script::builder::StructuredScript as Script;
 pub use bitcoin_script::script;
 
 use crate::bigint::U256;
-
 use crate::hash::blake3_utils::{compress, get_flags_for_block, TablesVars};
 
 // This implementation assumes you have the input is in compact form on the stack.
@@ -320,28 +317,28 @@ pub fn blake3_verify_output_script(expected_output: [u8; 32]) -> Script {
     }
 }
 
-#[cfg(any(feature = "fuzzing", test))]
-pub fn verify_blake_output(message: &[u8], expected_hash: [u8; 32]) {
-    use crate::execute_script_buf;
-    use bitcoin_script_stack::optimizer;
-
-    let mut bytes = blake3_push_message_script(&message).compile().to_bytes();
-    let optimized = optimizer::optimize(blake3_compute_script(message.len()).compile());
-    bytes.extend(optimized.to_bytes());
-    bytes.extend(
-        blake3_verify_output_script(expected_hash)
-            .compile()
-            .to_bytes(),
-    );
-    let script = ScriptBuf::from_bytes(bytes);
-    assert!(execute_script_buf(script).success);
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::execute_script_buf;
+    use bitcoin::ScriptBuf;
     use bitcoin_script_stack::optimizer;
+
+    pub fn verify_blake_output(message: &[u8], expected_hash: [u8; 32]) {
+        use crate::execute_script_buf;
+        use bitcoin_script_stack::optimizer;
+
+        let mut bytes = blake3_push_message_script(&message).compile().to_bytes();
+        let optimized = optimizer::optimize(blake3_compute_script(message.len()).compile());
+        bytes.extend(optimized.to_bytes());
+        bytes.extend(
+            blake3_verify_output_script(expected_hash)
+                .compile()
+                .to_bytes(),
+        );
+        let script = ScriptBuf::from_bytes(bytes);
+        assert!(execute_script_buf(script).success);
+    }
 
     fn verify_blake_outputs_cached<const LEN: usize>(
         messages: &[[u8; LEN]],
--- a/bitvm/src/hash/blake3.rs
+++ b/bitvm/src/hash/blake3.rs
@@ -321,11 +321,9 @@ pub fn blake3_verify_output_script(expected_output: [u8; 32]) -> Script {
 }
 
 #[cfg(any(feature = "fuzzing", test))]
-pub fn test_blake3_givenbyteslice(message: &[u8]) -> String {
+pub fn verify_blake_output(message: &[u8], expected_hash: [u8; 32]) {
     use crate::execute_script_buf;
     use bitcoin_script_stack::optimizer;
-
-    let expected_hash = blake3::hash(&message).as_bytes().clone();
 
     let mut bytes = blake3_push_message_script(&message).compile().to_bytes();
     let optimized = optimizer::optimize(blake3_compute_script(message.len()).compile());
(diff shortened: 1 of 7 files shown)


bitvm/src/hash/blake3.rs
```diff
@@ -235,416 +235,215 @@ pub fn blake3(
...
 
-    //change the hash representation from nibbles to bytes and compare with expected hash value
-    stack.custom(script!(
-        for (i, byte) in expected_hash.as_bytes().iter().enumerate(){
-            {*byte}
-            if i % 32 == 31{
+/// Returns a script that verifies the BLAKE3 output on the stack.
+///
+/// The script pops the BLAKE3 output and compares it with the given, expected output.
+pub fn blake3_verify_output_script(expected_output: [u8; 32]) -> Script {
```
AaronFeickert, line 308: Why not accept an array reference here? This would be more flexible for the caller.

fuzz/fuzz_targets/blake3.rs
```diff
@@ -1,28 +1,28 @@
...
-    }
-}
+static BLAKE3_COMPUTE_SCRIPT: LazyLock<ScriptBuf> = LazyLock::new(|| {
+    optimizer::optimize(blake3_compute_script(32).compile())
+});
 
-fuzz_target!(|data: LimitedBytes| {
-    // use full tables
-    test_blake3_givenbyteslice(&data.0, true);
+fuzz_target!(|message: [u8; 32]| {
```
AaronFeickert, line 16: Why target only 32-byte messages?

AaronFeickert (changes requested):

bitvm/src/hash/blake3.rs
uncomputable: Blake3 hashes are always 32-byte values. There is no reason to accept slices.

uncomputable: There is actually a movement in rust-bitcoin and rust-secp to replace slice references with arrays, when possible. It more clearly communicated which values are possible.

fuzz/fuzz_targets/blake3.rs
uncomputable: Each byte length necessitates a different BLAKE script. Precomputing even 10 of these takes considerable time. Fuzzing 32-byte inputs should give us confidence that the script works as intended. (See the commit message)

I can extend the fuzz target to cover all byte lengths used in BitVM
```bash
$ git grep 'hash_n_bytes'
bitvm/src/chunk/wrap_hasher.rs:pub(crate) fn hash_n_bytes<const N: u32>() -> Script {
bitvm/src/chunk/wrap_hasher.rs:    use crate::{bn254::{fp254impl::Fp254Impl, fq::Fq, fq2::Fq2}, chunk::{helpers::{pack_nibbles_to_limbs}, wrap_hasher::hash_n_bytes}, treepp::Script};
bitvm/src/chunk/wrap_hasher.rs:            { hash_n_bytes::<64>() }
bitvm/src/chunk/wrap_hasher.rs:            { hash_n_bytes::<128>() }
bitvm/src/chunk/wrap_hasher.rs:            {hash_n_bytes::<192>()}
bitvm/src/chunk/wrap_hasher.rs:            {hash_n_bytes::<448>()}
```

uncomputable:
Rebased onto `main` (#252)

bitvm/src/hash/blake3.rs
AaronFeickert: Right, but you can type it `&[u8; 32]`, right? Then you can ensure the required length while giving the caller the freedom to avoid clones, perhaps.

fuzz/fuzz_targets/blake3.rs
AaronFeickert: Agreed that fuzzing the lengths used in practice would be beneficial. Then we can have improved confidence that the fuzzer isn't relying on behavior specific to 32-byte messages.

AaronFeickert: > See the commit message

An interesting counterpoint to this is the SHA256 implementation(s). When fuzz testing was added in #239, it uncovered failures that only arose at non-obvious input lengths within the supported range.

uncomputable:
Rebased onto `main` and pushed two more commits. Undrafted the PR.

bitvm/src/hash/blake3.rs
uncomputable: As discussed privately, types that implement `Copy` should be passed by value. A function that takes an owned value has more flexibility in what it can do with it. For example, the script that we construct in `blake3_verify_output_script` iterates over the expected output bytes. The version on `main` uses `*byte` to deference each byte. My version can directly use `byte` because it has an owned array.

fuzz/fuzz_targets/blake3.rs
uncomputable: I updated the fuzz target to cover all message lengths that are used in BitVM. We already test length 0, 1 and 1024 in unit tests. All in all, that should give us good coverage with reasonable computational effort.

AaronFeickert: I think that's a fine compromise.

fuzz/fuzz_targets/blake3.rs
```diff
@@ -1,28 +1,75 @@
...
-
 use bitvm::execute_script_buf;
 use bitvm::hash::blake3::{
     blake3_compute_script, blake3_push_message_script, blake3_verify_output_script,
 };
 
-static BLAKE3_COMPUTE_SCRIPT: LazyLock<ScriptBuf> =
-    LazyLock::new(|| optimizer::optimize(blake3_compute_script(32).compile()));
+static BLAKE3_COMPUTE_SCRIPTS: LazyLock<[ScriptBuf; 4]> =
+    LazyLock::new(|| [64, 128, 128, 448]
```
AaronFeickert, line 16: Third element of this array is wrong.

AaronFeickert (changes requested):

fuzz/fuzz_targets/blake3.rs
uncomputable: Thanks for pointing this out. This typo is so easy to miss. I pushed a fix.

AaronFeickert: The danger of magic numbers!

AaronFeickert (approved):

PR #260, opened 2025-03-14
Prune unused code

AaronFeickert:
This PR prunes unused code.


AaronFeickert:
Note that these are part of the public API, so it should be confirmed before merging that they are not being actively used.

manishbista28 (approved):

manishbista28:
Being arithmetic functions, these would have only been used by bitvm module.
Since they aren't used inside bitvm or anywhere else, looks good to remove.

PR #262, opened 2025-03-14
CI: Run large tests in their own thread

lucidLuckylee:
We switched to a faster machine for the CI but it has less memory so can't run all the tests in parallel.
Running some of the bigger tests on a single thread should fix it


AaronFeickert:
Curious if there's any noticeable difference using `cargo nextest` instead.

Issue #259, opened 2025-03-12
Clippy lints


uncomputable:
Currently clippy generates a ton of lints, even though we have clippy in CI.

Is there a reason for that? I am happy to write a PR that resolves the lints.

AaronFeickert:
The CI can be updated to fail on warnings. But the repository settings should also require passing CI for merges, or it becomes too easy to merge before failures reveal themselves.

uncomputable:
#264 made progress towards fixing this issue, but there remain many warnings outside of `bitvm/bitvm`.

AaronFeickert:
It was suggested by @storopoli to do automated linting initially, since reviewers can locally check this without too much hassle.

PR #263, opened 2025-03-14
Make BLAKE3 Limb Length Generic

just-erray:
This PR changes the compact BLAKE3 code, so that it can work with various limb lengths. Previously, it worked only with 29, since the underlying BigInt limb length for field values was 29. By changing only 4 lines inside the BLAKE3 function, this update improves usability for future and outside use (since using a limb length of 4 enables the script for the unpacked version) while also serving as a small optimization for other cases.

Note that this PR also includes some small optimizations. It is currently in draft status due to significant conflicts with an [open PR](https://github.com/BitVM/BitVM/pull/257). Once that PR is merged, this code will undergo substantial revisions.


uncomputable:
Thanks for waiting for #257 :) We should be able to merge that PR soon.

AaronFeickert:
Please rebase now that #257 is merged.

just-erray:
Since the two PRs were too conflicting, I reset this branch and reapplied the changes on top of it. I also cleaned up `lib.rs` a bit while I got my hands dirty, but nothing major. I left fuzz test only running with limb length 29, since other tests should be more than enough for the others. It should be ready to merge now.

bitvm/src/hash/blake3.rs
```diff
@@ -50,15 +56,20 @@ fn blake3(stack: &mut StackTracker, mut msg_len: u32, define_var: bool, use_full
...
         "msg length must be less than or equal to 1024 bytes"
     );
+    assert!(
+        (4..32).contains(&limb_len),
+        "limb length must be in the range [4, 32)"
+    );
 
     //number of msg blocks
-    let num_blocks = f64::ceil(msg_len as f64 / 64_f64) as u32;
+    let num_blocks = (msg_len + 63) / 64;
```
AaronFeickert, line 65: Should be able to use `div_ceil` here instead.

bitvm/src/hash/blake3.rs
```diff
@@ -50,15 +56,20 @@ fn blake3(stack: &mut StackTracker, mut msg_len: u32, define_var: bool, use_full
...
 
     // If the compact form of message is on stack but not associated with variable, convert it to StackVariable
     if define_var {
+        let limb_count = (256 + limb_len as u32 - 1) / limb_len as u32;
```
AaronFeickert, line 69: Replace with `div_ceil` for clarity of intent.

bitvm/src/hash/blake3.rs
```diff
@@ -19,7 +19,13 @@ use crate::hash::blake3_utils::{compress, get_flags_for_block, TablesVars};
 /// ## See
 ///
 /// [`blake3_compute_script`].
-fn blake3(stack: &mut StackTracker, mut msg_len: u32, define_var: bool, use_full_tables: bool) {
+fn blake3(
+    stack: &mut StackTracker,
+    mut msg_len: u32,
+    define_var: bool,
+    use_full_tables: bool,
+    limb_len: u8,
```
AaronFeickert, line 27: Any restrictions on the limb length here that you're assuming?

bitvm/src/hash/blake3.rs
```diff
@@ -184,7 +203,7 @@ fn blake3(stack: &mut StackTracker, mut msg_len: u32, define_var: bool, use_full
     tables.drop(stack);
 
     // get the result hash
-    stack.from_altstack_joined(8_u32 * 8, "blake3-hash");
+    stack.from_altstack_joined(8 as u32 * 8, "blake3-hash");
```
AaronFeickert, line 206: Seems unnecessary.

bitvm/src/hash/blake3.rs
```diff
@@ -266,44 +281,44 @@ pub fn blake3_push_message_script(message_bytes: &[u8]) -> Script {
...
     );
     let mut stack = StackTracker::new();
     let use_full_tables = true;
     let message_len = message_len as u32; // safety: message_len <= 1024 << u32::MAX
-    blake3(&mut stack, message_len, true, use_full_tables);
+    blake3(&mut stack, message_len, true, use_full_tables, limb_len);
     stack.get_script()
 }
 
+/// See [`blake3_compute_script`].
```
AaronFeickert, line 317: Please document explicitly that this simply calls the generic function with 29-bit limbs.

bitvm/src/hash/blake3.rs
```diff
@@ -328,24 +343,41 @@ pub fn blake3_verify_output_script(expected_output: [u8; 32]) -> Script {
...
 mod tests {
     use super::*;
-    use crate::execute_script_buf;
+    use crate::execute_script_buf_without_stack_limit;
     use bitcoin::ScriptBuf;
     use bitcoin_script_stack::optimizer;
 
+    const RUN_EXTENSIVE_TESTS: bool = false;
+
+    const ALL_POSSIBLE_LIMB_LENGTHS: [u8; 28] = [
```
AaronFeickert, line 352: This seems to imply there are particular limits to allowed limbs. This is not checked anywhere.

AaronFeickert: It's also not documented anywhere.

bitvm/src/hash/blake3.rs
```diff
@@ -328,24 +343,41 @@ pub fn blake3_verify_output_script(expected_output: [u8; 32]) -> Script {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::execute_script_buf;
+    use crate::execute_script_buf_without_stack_limit;
     use bitcoin::ScriptBuf;
     use bitcoin_script_stack::optimizer;
 
+    const RUN_EXTENSIVE_TESTS: bool = false;
```
AaronFeickert, line 350: Document this.

bitvm/src/lib.rs
```diff
@@ -117,13 +117,34 @@ impl fmt::Display for ExecuteInfo {
...
+    execute_script_buf_optional_stack_limit(script, true)
+}
+
+pub fn execute_script_without_stack_limit(script: treepp::Script) -> ExecuteInfo {
+    execute_script_buf_optional_stack_limit(script.compile(), false)
+}
+
+pub fn execute_script_buf_without_stack_limit(script: bitcoin::ScriptBuf) -> ExecuteInfo {
+    execute_script_buf_optional_stack_limit(script, false)
+}
```
AaronFeickert, line 133: Do these need to be public?

bitvm/src/lib.rs
```diff
@@ -117,13 +117,34 @@ impl fmt::Display for ExecuteInfo {
...
+
+/// Executing a script on stack without `MAX_STACK_SIZE` limit is only for testing purposes \
+/// Don't use in production without the stack limit (i.e. `stack_limit` set to false)
+pub fn execute_script_buf_optional_stack_limit(
```
AaronFeickert, line 137: Does this need to be public?

AaronFeickert (changes requested):

bitvm/src/hash/blake3.rs
```diff
@@ -41,7 +47,7 @@ fn blake3(stack: &mut StackTracker, mut msg_len: u32, define_var: bool, use_full
             0,
             "push empty string hash in nibble form",
         );
-        stack.define(8_u32 * 8, "blake3-hash");
+        stack.define(8 as u32 * 8, "blake3-hash");
```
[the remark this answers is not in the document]
just-erray: I think this is left since I started editing from an old PR, reverting it

bitvm/src/hash/blake3.rs
just-erray: Good idea

bitvm/src/hash/blake3.rs
just-erray: Good idea

bitvm/src/hash/blake3.rs
just-erray: It is documented and asserted in the code, but I will make the documentation more detailed.

bitvm/src/hash/blake3.rs
just-erray: Same as before

bitvm/src/hash/blake3.rs
just-erray: Good idea

bitvm/src/hash/blake3.rs
just-erray: It is checked with asserts and is documented, but I'll document the test options

bitvm/src/lib.rs
just-erray: Yes, one of them is used in BLAKE3 tests, and in general, they were created previously and used for extensive testing.

bitvm/src/lib.rs
just-erray: No, changing it

bitvm/src/hash/blake3.rs
AaronFeickert: We're standardizing on this in #264 as well.

bitvm/src/hash/blake3.rs
AaronFeickert: Yeah, this is exactly the sort of thing that should be clear in the docstring since it's part of the public API.

uncomputable (changes requested):
Since we don't use limb lengths other than 29 and we have an upcoming audit, can we keep the function signatures of the blake3 functions the same and introduce private functions with the more general limb parameter?

- `blake3_compute_script` + `blake3_compute_script_with_limbs`
- `blake3_push_message_script` + `blake3_push_message_script_with_limbs`
- `blake3_verify_output_script` + `blake3_verify_output_script_with_limbs`

(Feel free to choose a different naming scheme)

Also, can you introduce a separate function to unit test different limb lengths? Iterating over all possible limb lengths just makes CI run for much longer. Especially, please use the old method for `test_single_byte` and `test_official_test_vectors`. Those tests run 10 seconds in the old version, and they will run for minutes your generalized version.

bitvm/src/hash/blake3.rs
```diff
@@ -229,34 +248,60 @@ pub fn blake3_push_message_script(message_bytes: &[u8]) -> Script {
...
-/// The stack MUST contain only the message.
-/// Anything else MUST be moved to the alt stack.
-/// When hashing the empty message (length 0), the stack MUST be empty.
-///
-/// ## Message Format Requirement
+/// ## Parameters:
+/// - `msg_len`: Length of the message. (excluding the padding, number of bytes)
+/// - `define_var`: Set to false if the input on stack is already defined as `StackTracker` varibles.
+/// - `use_full_tables`: toggle if you want to use full precomputation table or only half tables. Full table is script efficient but uses more stack.
+/// - `limb_len`: Limb length (number of bits per element) that the input in the stack is packed, for example it is 29 for current field elements
```
uncomputable, line 294: Why do you change the format of the doc string so much?

uncomputable: I would keep the `blake3_compute_script` function exactly the same and add a new function `blake3_compute_script_with_limbs` that is more general. I would explain the differences in the docstring for `blake3_compute_script_with_limbs`.

just-erray: I agree that moving panics to the top of the behavior section is better—I unconsciously made that change since I started from an old commit. However, I didn't make any other changes besides additions and topic renamings to the docstring, so the format remains nearly the same. I don't think creating two separate docstrings makes much sense. The only differences are one line in the message format and two added panics. One should be able to get the points just as easily from this version.

just-erray:
> Since we don't use limb lengths other than 29 and we have an upcoming audit, can we keep the function signatures of the blake3 functions the same and introduce private functions with the more general limb parameter?

Renaming it to the first version seems sensible without making it private, since the point is them being usable from outside. 

> Also, can you introduce a separate function to unit test different limb lengths? Iterating over all possible limb lengths just makes CI run for much longer. Especially, please use the old method for test_single_byte and test_official_test_vectors. Those tests run 10 seconds in the old version, and they will run for minutes your generalized version.

It should only take 3 times as much currently, which only takes 15 seconds in my environment. Extensive tests will be rarely necessary and aren't activated for CI.

bitvm/src/hash/blake3.rs
just-erray: Moved panics to the top and reformatted docstring so it is more readable

bitvm/src/hash/blake3.rs
```diff
@@ -229,34 +248,62 @@ pub fn blake3_push_message_script(message_bytes: &[u8]) -> Script {
...
 /// resulting in higher stack efficiency and support for larger messages.
 ///
-/// The message length is the number of message bytes without padding.
+/// ## Parameters:
 ///
-/// ## Empty Stack Requirement
+/// - `msg_len`: Length of the message. (excluding the padding, number of bytes)
+/// - `define_var`: Set to false if the input on stack is already defined as `StackTracker` varibles.
+/// - `use_full_tables`: toggle if you want to use full precomputation table or only half tables. Full table is script efficient but uses more stack.
+/// - `limb_len`: Limb length (number of bits per element) that the input in the stack is packed, for example it is 29 for current field elements
```
uncomputable, line 295: `define_var` and `use_full_tables` are no longer parameters of this function.

uncomputable:
> Extensive tests will be rarely necessary and aren't activated for CI
The point of unit tests is to run them, no?

If a test takes too long, we should mark is as `#[ignored]`. For example, we could adapt the test vectors and single byte units to iterate over all limb sizes and mark these tests as `#[ignored]`. We can still run these tests via `cargo test --ignored` but they don't slow down the remaining CI.

bitvm/src/hash/blake3.rs
just-erray: 🤦, fixing it

just-erray:
>If a test takes too long, we should mark is as #[ignored]. For example, we could adapt the test vectors and single byte units to iterate over all limb sizes and mark these tests as #[ignored]. We can still run these tests via cargo test --ignored but they don't slow down the remaining CI.

This solution seems better, changing it

bitvm/src/hash/blake3.rs
uncomputable: I think it would be easier to audit the blake function that is used in bitvm if we keep the old docstring.

The new docstring is of course more general, but we don't use that inside bitvm (yet). That is why I would create a second docstring that explains the differences.

just-erray:
> This solution seems better, so I am changing it.

I couldn't find a better solution than duplicating each function with wrappers in this case, and this seems like an unsustainable and messy approach. If there's a better way to handle this, adding two variants of the tests with `#[ignore]` would definitely be the better option. If not, I think leaving it as a variable and allowing developers to enable it when something related to limb length changes would be the better approach. In both cases, I don't think there will be any significant differences between them, since developers will have control over running tests while editing the code.

uncomputable:
I will try to come up with a solution and post a gist.

uncomputable:
Feel free to take a look at this: https://gist.github.com/uncomputable/021d98fac2bae02ace0b9005f4ba3484

Run the ignored tests with `cargo test -- --ignored`.

just-erray:
This looks nice. I think editing it slightly so that normal tests run with useful limbs (since it adds at most 10 seconds to the total test time) and ignored tests run with all limb lengths should make it good to go.

uncomputable:
Feel free to arrange the code as you want. I mainly wanted to show that it's possible to ignore the expensive tests without duplicating code.

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -4,7 +4,6 @@ use crate::{
     signatures::wots_api,
```
uncomputable, line 4: Please run `clippy --all-targets` and fix the warnings in `blake3.rs`. #259 should fix the warnings in all other files.

uncomputable (approved):
ACK 2dcd8adb9e4e2bff208b41d091a0751790eab307 Looks good to me. Thanks for accommodating my feedback. I haven't run the tests because my laptop is too weak. @AaronFeickert

PR #264, opened 2025-03-14
Fix clippy

### LINKED ISSUE #259: Clippy lints

uncomputable:
Progress towards #259. There are still many warnings outside of `bitvm/bitvm`.

There were a lot of clippy lints. Especially when running `cargo clippy --all-targets`. This PR resolves all of them.

I tried to be as conservative as possible, keeping the general function structure the same, not deleting any types, etc. At some places, I allowed the lint instead of resolving it. These lines are searchable via `git grep` and can be fixed in follow-up PRs.

The [manual_div_ceil lint explanation](https://rust-lang.github.io/rust-clippy/master/index.html#manual_div_ceil) might help reviewers.

There is also the usage of `cfg_chunks_mut!` from `ark_std` that caused problems due to the missing `parallel` feature. I replaced that line with [the non-parallel function call, as by the macro](https://github.com/arkworks-rs/std/blob/b0cbdeb097b42f61880b3bee19e8dd37258d23a5/src/lib.rs#L179-L192).

```
warning: unexpected `cfg` condition value: `parallel`
   --> bitvm/src/bn254/ell_coeffs.rs:329:21
    |
329 |         let mut f = cfg_chunks_mut!(pairs, 4)
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `fuzzing`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `cfg_chunks_mut` crate for guidance on how handle this unexpected cfg
    = help: the macro `cfg_chunks_mut` may come from an old version of the `ark_std` crate, try updating your dependency with `cargo update -p ark_std`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `#[warn(unexpected_cfgs)]` on by default
    = note: this warning originates in the macro `cfg_chunks_mut` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### DIFF
--- a/bitvm/src/bn254/ell_coeffs.rs
+++ b/bitvm/src/bn254/ell_coeffs.rs
@@ -6,7 +6,6 @@ use ark_ec::bn::g2::G2Prepared as ark_G2Prepared;
 use ark_ec::bn::{BnConfig, TwistType};
 use ark_ec::pairing::{MillerLoopOutput, Pairing, PairingOutput};
 use ark_ec::short_weierstrass::Affine;
-use ark_ec::short_weierstrass::SWCurveConfig;
 use ark_ec::{AffineRepr, CurveGroup};
 use ark_ff::Field;
 use ark_ff::{AdditiveGroup, CyclotomicMultSubgroup};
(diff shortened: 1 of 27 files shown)


AaronFeickert (approved):
LGTM, pending a rebase since #257 has been merged.

manishbista28 (approved):

manishbista28:
looks good to me

bitvm/src/bn254/ell_coeffs.rs
```diff
@@ -28,13 +27,15 @@ pub struct G2Prepared {
 pub type EllCoeff = (ark_bn254::Fq2, ark_bn254::Fq2, ark_bn254::Fq2);
 
 #[derive(Clone, Copy, Debug)]
+#[allow(dead_code)]
```
manishbista28, line 30: can remove G2HomProjective if it's in the scope of this PR

bitvm/src/chunk/g16_runner_core.rs
```diff
@@ -326,10 +326,10 @@ pub(crate) fn groth16_generate_segments(
...
 
     // (t2, t3) = (le.t2, le.t3);
-    let tmp_q2f = frob_q_power(pubs.q2, -1);
-    t2 = (t2 + tmp_q2f).into_affine();
-    let tmp_q3f = frob_q_power(pubs.q3, -1);
-    t3 = (t3 + tmp_q3f).into_affine();
+    // let tmp_q2f = frob_q_power(pubs.q2, -1);
+    // t2 = (t2 + tmp_q2f).into_affine();
+    // let tmp_q3f = frob_q_power(pubs.q3, -1);
+    // t3 = (t3 + tmp_q3f).into_affine();
```
manishbista28, line 332: can delete these comments too, but i'm okay either way.

AaronFeickert: Thanks @manishbista28 for checking this. I recommend deleting instead of commenting. We have them in the repository history if needed.

AaronFeickert: Same with lines 1058-1062 in the same file?

manishbista28: yes, same for comments in lines 1058-1062.

AaronFeickert:
We should update CI to fail on linter warnings (and enforce passing CI for merges).

uncomputable:
I removed the unused code and updated CI to deny clippy warnings.

AaronFeickert:
Many more lints from bridge.

uncomputable:
I reverted the CI change because clippy finds a lot of warnings in the bridge and in other places. Fixing these warnings is out of scope for this PR.

AaronFeickert (approved):
LGTM, pending CI.

AaronFeickert (approved):

uncomputable:
Rebased onto `main`

Issue #267, opened 2025-03-18
Issue when saving to remote data storage, error not being handled


kristiann57:
## ERROR Description:

When executing the “initiate peg-in” command, if bitVM encounters issues while saving data to the remote storage (AWS S3, FTP, FTPS etc…), it will continue to broadcast the transaction even if the data hasn't been saved correctly. This may lead to discrepancies, since other instances of bitVM (other verifiers, operators, etc.) will rely on that information to perform other relevant operations.

```
kritian@:~/demos/BitVM$ ./target/release/bridge initiate-peg-in --utxo ee29855315760b5b839ad20c9ce19a1e235c54afc2431b2a527b97458c0ab8e5:1 --destinat
ion_address 0x8C6bF4b04363910443cCc8F3B71B267EC3b96241
Using data file path: bridge_data/Operator 0/testnet/ethereum_sepolia/031ba4854cef9979134eb93ead94caf049af97974817b1b65448b94b620353f11b
Writing data file to 1741653513934-bridge-client-data-demo-feb-2025.json (size: 1134)
**Failed to push: Failed to save data file**
Created peg-in graph with ID: 7E2785CCC8BCA6422050E2BA2AE35FE76D293205AA2DCBEDFC93EEB6D265C740
Broadcasting deposit...
Tx broadcasted successfully. Txid: 27b80f3a93cab6206f9e3f9040803a53341203b27fcb1f019030e5f6e5be3e5
```

## **What's being saved?**

The BitVM client is trying to save the **`BitVMClientPublicData`** structure, which contains:

1. Version number
2. Peg-in graphs - critical data about peg-in transactions
3. Peg-out graphs - critical data about peg-out transactions

This data is essential for the proper functioning of the BitVM bridge, as it maintains the state of all transactions and allows different nodes to stay in sync.

## **Where is it being saved?**

The BitVM client attempts to save data in two places:

1.**Remote Data Store** - The system supports multiple remote backends:

- ◦ AWS S3
- ◦ FTP
- ◦ FTPS
- ◦ SFTP

The client attempts to initialize these backends during startup by checking for environment variables with credentials. If no credentials are found for any backend, operations like saving data files will fail with "Failed to save data file" errors.

2.**Local File System** - Even if the remote save fails, the client tries to save a local copy at:

```
CopyInsert
bridge_data/{user_prefix}/{source_network}/{destination_network}/{public_key}

```

## **Whereis the Error happening?**

[...]

PR #268, opened 2025-03-19
Fix error handling in flush() to prevent transaction broadcasting after storage failure

kristiann57:
This is a PR addressing [issue 267](https://github.com/BitVM/BitVM/issues/267)
##Problem
I noticed that when the flush() method in BitVMClient encounters an error while saving data to remote storage, the execution continues and attempts to broadcast transactions. This could lead to inconsistent state. 

I've made the following changes:
Modified flush() in client.rs to return a Result<(), String> instead of ()
Updated save_to_data_store() to properly propagate error messages
Changed the error handling in handle_initiate_peg_in_command() to stop execution if flush() fails

### DIFF
--- a/bridge/src/client/client.rs
+++ b/bridge/src/client/client.rs
@@ -244,8 +244,8 @@ impl BitVMClient {
         self.read_from_l2().await;
     }
 
-    pub async fn flush(&mut self) {
-        self.save_to_data_store().await;
+    pub async fn flush(&mut self) -> Result<(), String> {
+        self.save_to_data_store().await
     }
 
     /*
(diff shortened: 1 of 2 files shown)


AaronFeickert (commented):
utACK. Seems fine to me, but would recommend an additional reviewer with more familiarity with this aspect of the codebase.

bridge/src/client/client.rs
```diff
@@ -491,7 +491,7 @@ impl BitVMClient {
         (None, 0, 0)
     }
 
-    async fn save_to_data_store(&mut self) {
+    async fn save_to_data_store(&mut self) -> Result<(), String> {
         // read newly created data before pushing
         let latest_file_names_result = Self::get_latest_file_names(
```
eigmax, line 496: we can do like this:

if let Some(latest_file_names) = Self::get....).await {
            let mut latest_file_names = latest_file_names_result.unwrap();
            latest_file_names.reverse();
            let latest_valid_file_name = Self::process_files(self, latest_file_names).await;
            self.latest_processed_file_name = latest_valid_file_name;
}

eigmax: same for write_compressed_data.

eigmax: Same for write_compressed_data. 

and return an error at the end of this function.

PR #278, opened 2025-04-07
chore: fix array out-of-bounds error with index validation

barajeel:
added a validation check to prevent array out-of-bounds errors. now, if the index is incorrect, the code won't panic and will handle it more gracefully.

p.s.this should improve stability and avoid unexpected crashes.

### DIFF
--- a/header-chain/src/mmr_guest.rs
+++ b/header-chain/src/mmr_guest.rs
@@ -47,10 +47,6 @@ impl MMRGuest {
         }
         println!("GUEST: calculated subroot: {:?}", current_hash);
         println!("GUEST: subroots: {:?}", self.subroots);
-        if mmr_proof.subroot_idx < self.subroots.len() {
-            self.subroots[mmr_proof.subroot_idx] == current_hash
-        } else {
-            false
-        }
+        self.subroots.get(mmr_proof.subroot_idx) == Some(current_hash)
     }
 }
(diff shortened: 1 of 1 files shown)


header-chain/src/mmr_guest.rs
```diff
@@ -47,6 +47,10 @@ impl MMRGuest {
         }
         println!("GUEST: calculated subroot: {:?}", current_hash);
         println!("GUEST: subroots: {:?}", self.subroots);
-        self.subroots[mmr_proof.subroot_idx] == current_hash
+        if mmr_proof.subroot_idx < self.subroots.len() {
+            self.subroots[mmr_proof.subroot_idx] == current_hash
+        } else {
+            false
+        }
```
AaronFeickert, line 54: It's probably cleaner to use [`get`](https://doc.rust-lang.org/std/primitive.slice.html#method.get) instead.

AaronFeickert (changes requested):

header-chain/src/mmr_guest.rs
AaronFeickert: Also, is it clear that this case can actually arise?

AaronFeickert:
Have you checked whether or not the out-of-bounds case can arise?

barajeel:
Yeah, I looked into it — the index is safe with the current setup. I just used `get` as a precaution in case something changes later and messes with the index logic.

barajeel:
[AaronFeickert](https://github.com/AaronFeickert) yeah, I looked into it — the index is safe with the current setup. I just used get as a precaution in case something changes later and messes with the index logic.

AaronFeickert (approved):

AaronFeickert (changes requested):

AaronFeickert (approved):

AaronFeickert:
Approved, pending CI.

AaronFeickert:
@lucidLuckylee: Can you approve CI?

PR #280, opened 2025-04-15
perf(chunk): replace Script with ScriptBuf in batch operations

manishbista28:
This PR  replaces use of Vec<Script> with use of Vec\<ScriptBuf>

Reason:
chunk module's APIs accept, works with or return array of Script.

For example, while generating chunked scripts in the presigning stage, one of the functions you'd call is
`pub fn api_generate_partial_script(vk: &ark_groth16::VerifyingKey<Bn254>) -> Vec<Script>`

Holding data as Script in memory seems really costly compared to holding its equivalent form in ScriptBuf.
This is especially true when you're working with large batch of scripts like the function above.
From benchmarks, it seems Vec<Script> consumes around 10 times as much memory (RAM) compared to its serialized form Vec\<ScriptBuf>.

Changes are limited to type conversion only but has been done on public APIs.

### DIFF
--- a/bitvm/src/chunk/api.rs
+++ b/bitvm/src/chunk/api.rs
@@ -11,6 +11,7 @@ use crate::signatures::wots_api::{wots256, wots_hash};
 use crate::treepp::*;
 use ark_bn254::Bn254;
 use ark_ec::bn::Bn;
+use bitcoin::ScriptBuf;
 
 use super::api_runtime_utils::{
     execute_script_from_assertion, get_pubkeys, get_signature_from_assertion,
@@ -175,7 +176,7 @@ pub mod type_conversion_utils {
 // Step 1
 // The function takes public parameters (here verifying key) and generates partial script
 // partial script is essentially disprove script minus the bitcommitment locking script
-pub fn api_generate_partial_script(vk: &ark_groth16::VerifyingKey<Bn254>) -> Vec<Script> {
+pub fn api_generate_partial_script(vk: &ark_groth16::VerifyingKey<Bn254>) -> Vec<ScriptBuf> {
     generate_partial_script(vk)
 }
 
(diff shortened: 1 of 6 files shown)


Hakkush-07:
Better than my script lines approach for sure, but there is still one "script_lines" left in Fq6::hinted_square. I think we can get rid of it too in this PR. With that LGTM

AaronFeickert:
Why the huge difference, do you know?

lucidLuckylee (approved):

lucidLuckylee:
> Why the huge difference, do you know?

Note that `Script` is not the `bitcoin::Script` type but from https://github.com/BitVM/rust-bitcoin-script.
In rust-bitcoin-script there is still a lot of overhead due to the initial effort of an automatic chunker (wanted to preserve the call structure of scripts). That is basically the main reason why it uses up so much more memory than necessary.

The lib is currently quite cleaned up so if anyone wants to improve speed one can do so by getting rid of or speeding up this `script_map` in here: https://github.com/BitVM/rust-bitcoin-script/blob/519088c7e5e89769910dabcb1b663fa48a21a645/src/builder.rs#L33

AaronFeickert:
@lucidLuckylee thanks for the explanation! I was very confused on the assumption this was `bitcoin::Script`.

AaronFeickert (approved):
Approved, pending any changes suggested by @Hakkush-07.

lucidLuckylee:
I had AI update the static_script_map branch so if you are interested in performance gains you may be able to use it (iirc there were some issues with the multi-threading in cargo test but maybe they are resolved now)

AaronFeickert:
@lucidLuckylee is it just the latest commit on the branch that's relevant here?

lucidLuckylee:
yes

Hakkush-07:
I can open my changes as a separate PR right after this.

PR #282, opened 2025-04-16
Fix connector_b_leaf_2_script test for proper Taproot script verifica…

VolodymyrBg:
…tion

Configure test block headers for successful script verification in test_connector_b_leaf_2_script:
- Create committed_sb with lower weight (higher target bits)
- Set up disprove_sb with higher weight (lower target bits)
- Configure block times to satisfy conditions: start_time < disprove_sb.time < start_time + 2 weeks
- Reuse start_time for signature creation

### DIFF
--- a/bridge/src/connectors/connector_b.rs
+++ b/bridge/src/connectors/connector_b.rs
@@ -224,10 +224,22 @@ mod tests {
     fn test_connector_b_leaf_2_script() {
         const TWO_WEEKS_IN_SECONDS: u32 = 60 * 60 * 24 * 14;
 
-        // TODO: setup the test headers appropriately for the verification in the script to pass
-        let committed_sb = get_superblock_header();
+        // Setup test headers appropriately for the verification in the script to pass
+        let start_time = get_start_time_block_number(Regtest);
+        
+        // Create committed superblock with lower weight (higher hash value)
+        let mut committed_sb = get_superblock_header();
+        committed_sb.bits = CompactTarget::from_hex("0x1a030ecd").unwrap(); // Higher difficulty target
+        
+        // Create disprove superblock with higher weight (lower hash value)
         let mut disprove_sb = get_superblock_header();
-        disprove_sb.time = get_start_time_block_number(Regtest) + 1;
+        disprove_sb.bits = CompactTarget::from_hex("0x17030ecd").unwrap(); // Lower difficulty target
+        
+        // Set times to satisfy time constraints:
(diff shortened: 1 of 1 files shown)


AaronFeickert (changes requested):
The test still fails. Note that it's marked `#[ignore]`, so CI won't catch the failure.

VolodymyrBg:
> The test still fails. Note that it's marked `#[ignore]`, so CI won't catch the failure.

corrected as it seems to me

bridge/src/connectors/connector_b.rs
```diff
@@ -220,14 +220,61 @@ mod tests {
...
+        
+        // Serialize hashes to bytes and compare lexicographically
+        let committed_hash_bytes = committed_hash.to_string();
+        let disprove_hash_bytes = disprove_hash.to_string();
+        let is_disprove_heavier = disprove_hash_bytes < committed_hash_bytes;
+        println!("Is disprove SB heavier: {} (should be true)", is_disprove_heavier);
+        
+        if !is_disprove_heavier {
+            panic!("Disprove superblock is not heavier than committed superblock! Test precondition failed.");
+        }
```
AaronFeickert, line 276: You can directly compare the hashes without the need for serialization: `assert!(disprove_hash < committed_hash)`

This approach doesn't rely on the particulars of the serialization process.

AaronFeickert (changes requested):
One minor suggestion relating to serialization.

bridge/src/connectors/connector_b.rs
VolodymyrBg: > You can directly compare the hashes without the need for serialization: `assert!(disprove_hash < committed_hash)`
> 
> This approach doesn't rely on the particulars of the serialization process.

Done

AaronFeickert:
@lucidLuckylee: Can you approve CI?

AaronFeickert (approved):
LGTM, pending CI.

VolodymyrBg:
> LGTM, pending CI.

Hi, is this PR useful, or I should change something or close it if it's not actual anymore?

AaronFeickert:
@lucidLuckylee: Can you merge if we're not waiting for additional reviews?

AaronFeickert:
@VolodymyrBg: Out of curiosity, was this code generated by an AI system?

AaronFeickert:
No worries, just curious! Thanks for the contribution.

PR #286, opened 2025-04-21
Update chunk_instructions.md

leonarddt05:
Hello,

Here are some typos and corrections for this doc:

1. In "The following show tests for step wise execution"

"show" should be "shows",

2. In "function to generate execute chunked groth16 verifier and collect necessary information"

"generate execute" is likely a typo. Should be "generate and execute",

3. In "Because le.c1 has zero as the third coefficient, you get some necessary saving in terms of script and stack use"

"necessary saving" should be "savings" (plural).

Thanks.

### DIFF
--- a/docs/chunk_instructions.md
+++ b/docs/chunk_instructions.md
@@ -1,7 +1,7 @@
 To test Full Execution:
 cargo test --package bitvm --lib -- chunk::api::test::full_e2e_execution --exact --nocapture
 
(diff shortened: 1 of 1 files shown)


AaronFeickert:
The maintainers have previously indicated they may not merge small typo-fixing PRs, in part due to the possibility of airdrop-related incentives.

PR #289, opened 2025-04-30
WOTS data format refactor

uncomputable:
This PR documents the data formats used in the WOTS code. I also do some cheap refactors while I'm at it.

##  From messages to digits

In WOTS, we sign messages of bytes. These bytes are converted into digits, which reverses the byte order. Each byte is split into two nibbles (digits).

A message of the form `[0x12, 0x34, 0x56, 0x78, ... ]` is converted into digits `[..., 7, 8, 5, 6, 3, 4, 1, 2]`.

A checksum is appended to the end of the digit sequence.

This data format might raise some eyebrows, but a lot of code depends on it. In particular, many Bitcoin scripts expect this exact format. For now, it seems best to keep it and document as best as possible.

## From digits to signatures

A WOTS signature are the hashes of each digit, plus optionally the digit itself. The precise format is documented in the code.

### DIFF
--- a/bitvm/src/signatures/utils.rs
+++ b/bitvm/src/signatures/utils.rs
@@ -11,64 +11,90 @@ pub(super) const fn log_base_ceil(n: u32, base: u32) -> u32 {
     res
 }
 
-/// Converts the number to given base, smallest digit being at the start (returns `digit_count` smallest digits of it, and standard base representation of it if `digit_count` = -1)
-pub(super) fn to_digits(mut number: u32, base: u32, digit_count: i32) -> Vec<u32> {
-    let mut digits = Vec::new();
-    if digit_count == -1 {
-        while number > 0 {
-            let digit = number % base;
-            number = (number - digit) / base;
-            digits.push(digit);
-        }
-    } else {
-        digits.reserve(digit_count as usize);
-        for _ in 0..digit_count {
-            let digit = number % base;
-            number = (number - digit) / base;
-            digits.push(digit);
-        }
+/// Converts the given `checksum` into a vector of digits.
+///
+/// ## Output format
+///
+/// - sequence of `n_digits` many digits
+/// - each digit a `u32` value in range `0..2.pow(log2_base)`
+/// - checksum converted into BE bytes, in turn converted into digits
+pub(super) fn checksum_to_digits(mut checksum: u32, base: u32, n_digits: u32) -> Vec<u32> {
+    debug_assert!((16..=256).contains(&base));
+    debug_assert!(
+        base.checked_pow(n_digits)
+            .map(|upper_limit| checksum < upper_limit)
+            .unwrap_or(true),
+        "Checksum is too large to fit into the given number of digits"
+    );
+
+    let mut digits = vec![0; n_digits as usize]; // cast safety: 32-bit machine or higher
+
+    for digit in digits.iter_mut().rev() {
+        *digit = checksum % base;
+        checksum = (checksum - *digit) / base;
     }
+
     digits
 }
 
-/// Converts the given bytes to 'len' `u32`'s, each consisting of given number of bits
-pub(crate) fn bytes_to_u32s(len: u32, bits_per_item: u32, bytes: &[u8]) -> Vec<u32> {
-    assert!(
-        bytes.len() as u32 * 8 <= len * bits_per_item,
-        "Message length is too large for the given length"
+/// Converts the given `message` into a vector of digits.
+///
+/// ## Output format
+///
+/// - sequence of `n_digits` many digits
+/// - each digit a `u32` value in range `0..2.pow(log2_base)`
+/// - message bytes are reversed (but not their nibbles!)
+pub(crate) fn message_to_digits(n_digits: u32, log2_base: u32, message: &[u8]) -> Vec<u32> {
+    debug_assert!((4..=8).contains(&log2_base));
+    debug_assert!(
+        message.len() as u32 * 8 <= n_digits * log2_base,
+        "Message is too long to fit into the given number of digits"
     );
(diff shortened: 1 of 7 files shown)


bitvm/src/signatures/utils.rs
```diff
@@ -11,64 +11,95 @@ pub(super) const fn log_base_ceil(n: u32, base: u32) -> u32 {
...
-            number = (number - digit) / base;
-            digits.push(digit);
-        }
+/// Converts the given `checksum` into a vector of digits.
+///
+/// There are `n_digits` many digits.
+/// Each digit is a `u32` value in the range `0..base`.
+/// Digits are in big endian order.
+///
+/// ## Panics
```
AaronFeickert, line 20: Not in release mode.

bitvm/src/signatures/utils.rs
```diff
@@ -11,64 +11,95 @@ pub(super) const fn log_base_ceil(n: u32, base: u32) -> u32 {
...
-    assert!(
-        bytes.len() as u32 * 8 <= len * bits_per_item,
-        "Message length is too large for the given length"
+/// Converts the given `message` into a vector of digits.
+///
+/// There are `n_digits` many digits.
+/// Each digit is a `u32` value in the range `0..2.pow(log2_base)`.
+/// Digits are in big endian order.
+///
+/// ## Panics
```
AaronFeickert, line 47: Not in release mode.

bitvm/src/signatures/utils.rs
uncomputable: I changed an `assert!` into a `debug_assert!`. Thanks for the headsup.

uncomputable:
I documented most methods, focusing on the used byte formats and script inputs / outputs. The last commit renames "blocks" to "digits" for term consistency.

The WOTS API is still very general, as digits can be 5 bits, 6 bits or 7 bits long. This makes it hard to write documentation, because these corner cases are weird. I plan a follow-up PR that makes these general WOTS implementations private.

storopoli (changes requested):
I'd love to see also some JSON test vectors somewhere so that we can have an official agnostic test vectors for BitVM WOTS API.

bitvm/src/signatures/winternitz.rs
```diff
@@ -11,68 +11,61 @@ type HashOut = [u8; 20];
...
-    pub const fn new(message_block_count: u32, block_length: u32) -> Self {
+    /// Creates parameters for messages of the given number of digits of the given base.
+    ///
+    /// The base must be in the range `16..=256`.
+    pub const fn new(message_digit_len: u32, log2_base: u32) -> Self {
         assert!(
-            4 <= block_length && block_length <= 8,
-            "You can only choose block lengths in the range [4, 8]"
+            4 <= log2_base && log2_base <= 8,
+            "log2_base must be in the range 4..=8"
```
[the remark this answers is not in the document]
uncomputable: If the base is in the range `16..=256`, then log2(base) is in the range `4..=8`. So, technically, the docs are accurate. What is confusing is the difference between base and log2(base). Some methods take the former, some take the latter.

I think we should consistently use one over the other. The compiler should collapse `2.pow(log2_base)`, or `base.log2()` at compile time. We could also use min const generics.

uncomputable:
This PR focuses on the generic WOTS implementation. I plan to add test vectors in a follow-up PR for the 3 specific WOTS implementations that we use in BitVM.

bitvm/src/signatures/winternitz.rs
```diff
@@ -83,70 +76,105 @@ pub fn digit_signature(secret_key: &SecretKey, digit_index: u32, message_digit:
...
-    let mut public_key = PublicKey::with_capacity(ps.total_length() as usize);
-    for i in 0..ps.total_length() {
+    let mut public_key = PublicKey::with_capacity(ps.total_digit_len() as usize);
+    for i in 0..ps.total_digit_len() {
         public_key.push(public_key_for_digit(ps, secret_key, i));
     }
     public_key
 }
 
 /// Checksum of the message (negated sum of the digits)
```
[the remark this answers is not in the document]
uncomputable: I removed "negated sum of digits" from the docstring. How the checksum is computed is an implementation detail, and the function is private anyway.

bitvm/src/signatures/winternitz.rs
```diff
@@ -158,7 +186,6 @@ impl<VERIFIER: Verifier, CONVERTER: Converter> Default for Winternitz<VERIFIER,
     }
 }
 
-/// Implementation of the default functions regardless of the algorithms that are chosen
 impl<VERIFIER: Verifier, CONVERTER: Converter> Winternitz<VERIFIER, CONVERTER> {
     pub const fn new() -> Self {
```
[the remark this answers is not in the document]
uncomputable: I plan to make this function private in a follow-up PR.

storopoli (approved):
ACK ef11503e9d22d4b63c834c213f68a4604c69dbbb

bitvm/src/signatures/utils.rs
```diff
@@ -11,64 +11,90 @@ pub(super) const fn log_base_ceil(n: u32, base: u32) -> u32 {
...
-            let digit = number % base;
-            number = (number - digit) / base;
-            digits.push(digit);
-        }
+/// Converts the given `checksum` into a vector of digits.
+///
+/// ## Output format
+///
+/// - sequence of `n_digits` many digits
+/// - each digit a `u32` value in range `0..2.pow(log2_base)`
```
AaronFeickert, line 19: The value `log2_base` is not defined as part of the function signature.

AaronFeickert (commented):
As noted elsewhere, this PR would be a good opportunity to add JSON test vectors to aid other implementations.

PR #293, opened 2025-05-13
Updated WOTS API

### LINKED ISSUE #290: Make generic WOTS private
Seem to use three instantiations of WOTS in BitVM:

1. 16-byte messages split into 32 digits of 4-bit each (wots_hash)
2. 20-byte messages split into 40 digits of 4-bit each (blake3)
3. 32-byte messages split into 64 digits of 4-bit each (wots256)

I think we should have an API where you pick a variant and then you have the same functionality for each of them. That is, key management, signing and verification Bitcoin scripts.

Meanwhile, I think we should hide the generic functionality in `winternitz.rs`. We can make `Parameters` private and statically enforce good parameters for each variant.

### LINKED ISSUE #291: Add WOTS test vectors
The WOTS functionality should have accessible test vectors (_e.g._ in JSON).

uncomputable:
This PR introduces an API for using WOTS signatures, which enforces correct message and public key lengths. Test vectors are added that will help third party implementations verify compatibility with BitVM. The signature format remains unchanged. The (used parts) of the old API remain available.

The PR tries to touch as few lines as possible in the surrounding BitVM code, which turned out to be challenging, since WOTS is such a ubiquitous, low-level primitive. I was able to port the code in `bitvm/bitvm`. The code in `bitvm/bridge` remains mostly unchanged, as it would require a huge refactor to port, which I judged as out of scope.

Fixes #291 Fixes #290 

## Lessons learned

I learned a few lessons about WOTS while making this PR:

BitVM uses WOTS for many more message lengths than just 16 or 32 bytes. The superblock message is 80 bytes long, the destination network txid is 64 bytes long, the start time message is 4 bytes long, and so on. To be fair, most of this lives in `bitvm/bridge`.

The 20-byte WOTS variant for BLAKE3 is no longer used. We switched to 16-byte WOTS a while ago. This PR removes the old code.

Winternitz signatures consist of digit signatures (plus an optional digit). The digit signature is a byte sequence, which may have a trailing zero. The chance of this happening is 1/256. Bitcoin Core enforces minimal data pushes for standard transactions, which means that trailing zero bytes have to be removed. IDK exactly how this works, but trailing zeros might make some BitVM transactions non-standard.

[Secret keys consist of ASCII bytes of hex characters](https://github.com/BitVM/BitVM/blob/d3ee6062650448aef774966e82807d160b56a660/bitvm/src/signatures/signing_winternitz.rs#L34-L45). A random secret key is made from 20 random bytes, which are converted into 40 hex characters, which are converted into 40 ASCII bytes. A string is converted into a secret key by taking the ASCII bytes, which are converted into hex characters, which are in turn converted into (more) ASCII bytes. I'm not saying this is insecure, but it is confusing and it's good to be aware of this.

`Wots::checksig_verify_and_clear_stack` uses fewer drop operations than [winternitz_message_checksig_verify](https://github.com/BitVM/BitVM/blob/d3ee6062650448aef774966e82807d160b56a660/bitvm/src/signatures/signing_winternitz.rs#L120-L132). This could be fixed in a future PR, if someone wants to.


bitvm/src/chunk/api_runtime_utils.rs
```diff
@@ -552,8 +548,12 @@ pub(crate) fn execute_script_from_signature(
             for index in index_of_bitcommitted_msg {
                 let sig_data = &bitcom_sig_arr[index as usize];
                 let sig_preimage = match sig_data {
-                    SigData::SigHash(signature) => signature.to_compact_script(),
-                    SigData::Sig256(signature) => signature.to_compact_script(),
+                    SigData::SigHash(signature) => Wots16::compact_signature_to_raw_witness(
+                        &Wots16::signature_to_compact_signature(signature),
+                    ),
+                    SigData::Sig256(signature) => Wots32::compact_signature_to_raw_witness(
+                        &Wots32::signature_to_compact_signature(signature),
```
storopoli, line 555: Why do we have the mappping:

- `SigHash` -> `Wots16`
- `Sig256` -> `Wots32`

Wouldn't it better to have all of them `{Sig,Wots}{16,32}`?

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+    ///
+    /// For backwards compatibility, the original conversion function is used.
+    /// The `secret` string is converted into ASCII bytes,
+    /// which are in turn converted into lower hex ASCII bytes.
+    ///
+    /// Usage of this method is generally discouraged.
+    /// It is better to create a [`WinternitzSecret`] via [`Vec<u8>`].
+    fn secret_from_str(secret: &str) -> WinternitzSecret {
+        secret.as_bytes().to_lower_hex_string().into_bytes()
+    }
```
storopoli, line 72: Add a `deprecated` in this unutterable cursed ugly mess:

```rs
#[deprecated(since="X.Y.Z", note="please use `new_method` instead")]
```

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+/// Winternitz signatures for 32-byte messages.
+pub struct Wots32;
+/// Winternitz signatures for 64-byte messages.
+pub struct Wots64;
+/// Winternitz signatures for 72-byte messages.
+pub struct Wots72;
+/// Winternitz signatures for 80-byte messages.
+pub struct Wots80;
+
+macro_rules! impl_wots {
```
storopoli, line 338: docstring in the `impl_wots` macro

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+            secret_key,
+            public_key,
+            witness,
+            signature,
+            compact_witness,
+            compact_signature,
+        }
+    }
+
+    #[allow(dead_code)]
```
storopoli, line 651:
```suggestion
    #[expect(dead_code)]
```

If this thing is used, then this `expect` lint would fail.
I prefer that to `allow` which flies under the radar.

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+    fn extract_witness_from_stack(res: ExecuteInfo) -> Vec<Vec<u8>> {
+        res.final_stack.0.iter_str().fold(vec![], |mut vector, x| {
+            vector.push(x);
+            vector
+        })
+    }
+
+    /// Compare two elements of n length.
+    /// If them are not equal, return script's failure directly.
+    pub fn equalverify(n: usize) -> Script {
```
storopoli, line 402: are we using this fn outside this test module?
if not, make this non`pub`

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+            }
+
+            for _ in 0..n {
+                OP_FROMALTSTACK
+                OP_EQUALVERIFY
+            }
+        )
+    }
+
+    pub fn u32_witness_to_bytes(witness: Vec<Vec<u8>>) -> Vec<u8> {
```
storopoli, line 420: are we using this fn outside this test module?
if not, make this non`pub`

bitvm/src/signatures/signing_winternitz.rs
```diff
@@ -21,6 +20,18 @@ pub struct WinternitzSecret {
...
+    Winternitz::new();
+
+/// Winternitz verifier, returns the message in bytes
+pub const WINTERNITZ_VARIABLE_VERIFIER: Winternitz<ListpickVerifier, ToBytesConverter> =
+    Winternitz::new();
+
+/// Winternitz verifier for compact signature representation, returns the message in bytes
+pub const WINTERNITZ_MESSAGE_COMPACT_VERIFIER: Winternitz<BruteforceVerifier, VoidConverter> =
+    Winternitz::new();
+
```
storopoli, line 34: For the sake of consistency with other docstrings in this module:
```suggestion
/// Winternitz verifier, returns the message in digits.
pub const WINTERNITZ_MESSAGE_VERIFIER: Winternitz<ListpickVerifier, VoidConverter> =
    Winternitz::new();

/// Winternitz verifier, returns the message in bytes.
pub const WINTERNITZ_VARIABLE_VERIFIER: Winternitz<ListpickVerifier, ToBytesConverter> =
    Winternitz::new();

/// Winternitz verifier for compact signature representation, returns the message in bytes.
pub const WINTERNITZ_MESSAGE_COMPACT_VERIFIER: Winternitz<BruteforceVerifier, VoidConverter> =
    Winternitz::new();

```

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+        D: Deserializer<'de>,
+    {
+        let vec_s: Vec<String> = Vec::deserialize(deserializer)?;
+        vec_s
+            .into_iter()
+            .map(|s| Vec::<u8>::from_hex(&s).map_err(serde::de::Error::custom))
+            .collect()
+    }
+
+    #[allow(dead_code)]
```
storopoli, line 611:
```suggestion
    #[expect(dead_code)]
```

If this thing is used, then this `expect` lint would fail.
I prefer that to `allow` which flies under the radar.

storopoli (changes requested):
minor nits.
Amazing work! Thanks!

bitvm/src/signatures/public.rs
uncomputable: The method is actually not used in the code. I added it to make it easier to port the bridge code to the new API, which turned out to be too much work.

I can remove the method from `public.rs` and add a warning + deprecation to `signing_winternitz.rs`.

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+            .iter()
+            .map(|digit_sig| digit_sig[20])
+            // Remove the checksum at the end
+            .take(Self::PARAMETERS.message_digit_len as usize)
+            // Un-reverse digits
+            .rev()
+            .collect();
+        // Convert little endian digits [LSD, MSD] to big endian [MSD, LSD].
+        // "LSD" means "least significant digit" and
+        // "MSB" means "most significant digit".
```
AaronFeickert, line 176:
```suggestion
        // "MSD" means "most significant digit".
```

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+        }
+    };
+}
+
+impl_wots!(Wots4, 4, VoidConverter);
+impl_wots!(Wots16, 16, VoidConverter);
+impl_wots!(Wots32, 32, VoidConverter);
+impl_wots!(Wots64, 64, VoidConverter);
+impl_wots!(Wots72, 72, ToBytesConverter);
+impl_wots!(Wots80, 80, VoidConverter);
```
AaronFeickert, line 367: Is it worthwhile to include an explanatory comment to indicate the expected/existing use cases for these types?

bitvm/src/signatures/public.rs
AaronFeickert: Also please include at least brief internal documentation for this.

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+    use ark_std::test_rng;
+    use bitcoin::hex::{DisplayHex, FromHex};
+    use bitcoin::script::read_scriptint;
+    use bitcoin_script::script;
+    use rand::{RngCore as _, SeedableRng as _};
+    use serde::ser::SerializeSeq;
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+    use serde_json;
+
+    fn extract_witness_from_stack(res: ExecuteInfo) -> Vec<Vec<u8>> {
```
AaronFeickert, line 393: Please include at least brief internal documentation for this function.

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+    {
+        let vec_s: Vec<String> = Vec::deserialize(deserializer)?;
+        vec_s
+            .into_iter()
+            .map(|s| Vec::<u8>::from_hex(&s).map_err(serde::de::Error::custom))
+            .collect()
+    }
+
+    #[allow(dead_code)]
+    fn create_test_vector<WOTS: Wots + CompactWots>(
```
AaronFeickert, line 612: Please include at least brief internal documentation for this function.

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+            public_key,
+            witness,
+            signature,
+            compact_witness,
+            compact_signature,
+        }
+    }
+
+    #[allow(dead_code)]
+    fn generate_test_vectors_for<WOTS: Wots + CompactWots, const MSG_BYTE_LEN: usize>(
```
AaronFeickert, line 652: Please include at least brief internal documentation for this function.

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+        test_vectors.push(create_test_vector::<WOTS>(&message, secret_key));
+
+        test_vectors
+    }
+
+    // Uncomment and run this code to generate fresh test vectors:
+    // cargo test generate_and_save_test_vectors
+    /*
+    #[test]
+    fn generate_and_save_test_vectors() -> io::Result<()> {
```
AaronFeickert, line 725: Can this be flagged as ignored instead?

bitvm/src/signatures/public.rs
```diff
@@ -0,0 +1,838 @@
...
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+
+        let test_vectors: Vec<TestVector> = serde_json::from_str(&contents)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+
+        Ok(test_vectors)
+    }
+
+    fn verify_test_vector<WOTS: Wots + CompactWots>(test_vector: TestVector) {
```
AaronFeickert, line 754: Please include at least brief internal documentation for this.

bitvm/wots-test-vectors.json
AaronFeickert, line 1: Did not review this file.

bridge/src/commitments.rs
```diff
@@ -116,9 +116,9 @@ impl CommitmentMessageId {
             commitment_map.insert(
                 CommitmentMessageId::Groth16IntermediateValues((
                     format!("{}", i + NUM_PUBS + NUM_U256),
-                    wots_api::HASH_LEN as usize,
+                    16,
```
AaronFeickert, line 119: Can we use a constant here instead of a magic number?

bridge/src/commitments.rs
```diff
@@ -116,9 +116,9 @@ impl CommitmentMessageId {
...
                 )),
-                WinternitzSecret::new(wots_api::HASH_LEN as usize),
+                WinternitzSecret::new(16),
```
AaronFeickert, line 121: Can we use a constant here instead of a magic number?

bridge/tests/bridge/helper.rs
```diff
@@ -297,10 +296,7 @@ pub fn get_intermediate_variables_cached() -> BTreeMap<String, usize> {
             intermediate_variables.insert(format!("{}", i + NUM_PUBS), 32);
         }
         for i in 0..NUM_HASH {
-            intermediate_variables.insert(
-                format!("{}", i + NUM_PUBS + NUM_U256),
-                wots_api::HASH_LEN as usize,
-            );
+            intermediate_variables.insert(format!("{}", i + NUM_PUBS + NUM_U256), 16);
```
AaronFeickert, line 299: Can we use a constant here instead of a magic number?

AaronFeickert (changes requested):

bitvm/src/signatures/public.rs
uncomputable: There are already ignored tests that we expect to run via `cargo test -- --ignored`. If generating test vectors is an ignored test, then it will overwrite the existing test vectors. Someone with little knowledge of git may accidentally commit 6000 lines of changed JSON.

AaronFeickert: Got it. Would be nice to have this as a separate executable, so it would be subject to checks in the event of future changes. Is that reasonable to do?

uncomputable: I put the test vector generator next to the test vector verifier so it's easy to modify later. Adding a separate executable increases complexity, which seems unnecessary to me. I don't expect frequent updates to the test vectors.

bitvm/src/signatures/public.rs
uncomputable: I will make the `impl_wots!` macro public, so you can generate the WOTS implementations that you need downstream.

bitvm/src/signatures/public.rs
uncomputable: `#[expect(dead_code)]` complains that the actually dead code is supposedly not dead. I left `#[allow(dead_code)]` for now.

```
warning: this lint expectation is unfulfilled
   --> bitvm/src/signatures/public.rs:637:14
    |
637 |     #[expect(dead_code)]
    |              ^^^^^^^^^
    |
    = note: `#[warn(unfulfilled_lint_expectations)]` on by default
```

sander2:
> Winternitz signatures consist of digit signatures (plus an optional digit). The digit signature is a byte sequence, which may have a trailing zero. The chance of this happening is 1/256. Bitcoin Core enforces minimal data pushes for standard transactions, which means that trailing zero bytes have to be removed. IDK exactly how this works, but trailing zeros might make some BitVM transactions non-standard.

Are you sure this is the case? I never ran into this on regtest

uncomputable:
> Are you sure this is the case? I never ran into this on regtest. If that were true then that would be a huge problem not just for winternitz, but regular signature checks as well. I think maybe this only applies when the data is interpreted as an integer rather than a byte array

You are probably right. Bitcoin Script was weird rules around encodings. To be on the safe side, I added test vectors for leading and for trailing zero bytes.

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -2,14 +2,13 @@ use super::elements::ElementType;
...
     hash::blake3::blake3_compute_script,
-    signatures::wots_api,
     treepp::*,
 };
 use hash_utils::{
     hash_fp2, hash_fp6, hash_g2acc, hash_g2acc_with_hash_t, hash_g2acc_with_hashed_le,
 };
 
-pub const BLAKE3_HASH_LENGTH: usize = wots_api::HASH_LEN as usize;
+pub const BLAKE3_HASH_LENGTH: usize = 16;
```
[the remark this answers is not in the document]
uncomputable: _Don't make me recursively document the entire bitvm code base xD_

AaronFeickert (approved):
ACK 50bc2027c790492375df27b51b5c7d1cb6271f3f

storopoli (approved):
ACK 9941d769d04e24984f5b035a128587235c9564fe

AaronFeickert:
Please run the formatter, as it currently fails after recent changes.

storopoli (approved):
ACK bb0aff58dcde543d47bc7c0d64ca9479d40ac831

AaronFeickert:
Any idea why the guest build is failing CI?

uncomputable:
Looks like flaky CI to me. I have not touched the header chain circuit in this PR

PR #294, opened 2025-05-16
chore: cache hashing scripts

sander2:
Added some caching on the functions that create the hashing scripts to make running the tests a little bit more bearable. This reduces `generate_segments_using_mock_proof` execution time on my machine from 270s to 90s, a 4x improvement. The downside is increased memory usage, which I think is probably fine, but if there are objections we could make the caching conditional on a feature flag

### DIFF
--- a/bitvm/src/chunk/wrap_hasher.rs
+++ b/bitvm/src/chunk/wrap_hasher.rs
@@ -37,9 +37,11 @@ pub(crate) mod hash_utils {
         treepp::Script,
     };
     use bitcoin_script::script;
+    use cached::proc_macro::cached;
 
     /// Compute hash of top two field elements on stack: [a00, a01]
     /// Output is {BLAKE3_HASH_LENGTH} byte output represented in limb-form
+    #[cached]
     pub(crate) fn hash_fp2() -> Script {
         script! {
             // [a00, a01]
@@ -51,6 +53,7 @@ pub(crate) mod hash_utils {
 
     /// Compute hash of top four field elements on stack: [a00, a01, a10, a11]
     /// Output is {BLAKE3_HASH_LENGTH} byte output represented in limb-form
+    #[cached]
     pub(crate) fn hash_fp4() -> Script {
         script! {
             // [a00, a01, a10, a11]
@@ -65,6 +68,7 @@ pub(crate) mod hash_utils {
 
     /// Compute hash of top six field elements on stack: [a00, a01, a10, a11, a20, a21]
     /// Output is {BLAKE3_HASH_LENGTH} byte output represented in limb-form
+    #[cached]
     pub(crate) fn hash_fp6() -> Script {
         script! {
             // [a00, a01, a10, a11, a20, a21]
@@ -79,6 +83,7 @@ pub(crate) mod hash_utils {
 
     /// Compute hash of top six field elements on stack: [a00, a01,.., a60, a61]
     /// Output is {BLAKE3_HASH_LENGTH} byte output represented in limb-form
+    #[cached]
     pub(crate) fn hash_fp14() -> Script {
         script! {
             // [a00, a01,... ,a60, a61]
@@ -93,6 +98,7 @@ pub(crate) mod hash_utils {
 
     /// Compute hash of G2 Point Accumulator (i.e. [t(4), partial_product(14)]) where Hash(partial_product) has been passed as auxiliary input on stack.
     /// Stack: [t(4), Hash_partial_product(1)]
+    #[cached]
     pub(crate) fn hash_g2acc_with_hashed_le() -> Script {
         script! {
             // [t, Hash_partial_product]
@@ -108,6 +114,7 @@ pub(crate) mod hash_utils {
 
     /// Compute hash of G2 Point Accumulator (i.e. [t(4), partial_product(14)]) where all elements are passed as raw value on stack
     /// Stack: [t(4), partial_product(14)]
+    #[cached]
     pub(crate) fn hash_g2acc() -> Script {
         script! {
             // [t, partial_product]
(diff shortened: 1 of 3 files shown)


sander2:
No there isn't, it's just more readable this way. I can change it if you want though

AaronFeickert:
I don't have a particular preference; just unfamiliar with the inner workings of that crate.

AaronFeickert:
On further discussion, how much of a refactor would it be to use `OnceCell` or `LazyCell` for these test purposes, in order to remove the need for the extra dependency?

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -41,113 +42,148 @@ pub(crate) mod hash_utils {
...
+        CACHE
+            .get_or_init(|| {
+                script! {
+                    // [a00, a01]
+                    { hash_n_bytes::<64>() }
+                    // Hash(a00|a01)
+                    { pack_nibbles_to_limbs() }
+                }
+            })
+            .clone()
```
storopoli, line 55: Why do we need to `clone` inside the cache?

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -41,113 +42,148 @@ pub(crate) mod hash_utils {
...
+                    // [a00, a01, a10, a11]
+                    {Fq2::roll(2)}
+                    // [a10, a11, a00, a01]
+                    // Requires first msg-block at the top of stack
+                    { hash_n_bytes::<128>() }
+                    // Hash(a00|a01|a10|a11)
+                    { pack_nibbles_to_limbs() }
+                }
+            })
+            .clone()
```
storopoli, line 74: Why do we need to `clone` inside the cache?

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -41,113 +42,148 @@ pub(crate) mod hash_utils {
...
+                    // [a00, a01,... ,a60, a61]
+                    {Fq2::roll(2)} {Fq2::roll(4)} {Fq2::roll(6)}
+                    {Fq2::roll(8)} {Fq2::roll(10)} {Fq2::roll(12)}
+                    // [a60, a61,... ,a00, a01]
+                    {hash_n_bytes::<448>()}
+                    // Hash(a00|a01|..|a60|a61)
+                    {pack_nibbles_to_limbs()}
+                }
+            })
+            .clone()
```
storopoli, line 112: Why do we need to `clone` inside the cache?

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -41,113 +42,148 @@ pub(crate) mod hash_utils {
...
+                    {Fq::toaltstack()}
+                    {hash_fp4()}
+                    {Fq::fromaltstack()}
+                    // [Hash_t, Hash_partial_product]
+                    {hash_fp2()}
+                    // [ Hash(Hash_t|Hash_partial_product) ]
+                    // [ Hash(G2Acc) ]
+                }
+            })
+            .clone()
```
storopoli, line 132: Why do we need to `clone` inside the cache?

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -41,113 +42,148 @@ pub(crate) mod hash_utils {
...
+                    // [Hash_partial_product] [Hash_t]
+                    {Fq::fromaltstack()}
+                    {Fq::roll(1)}
+                    // [ Hash_t, Hash_partial_product ]
+                    {hash_fp2()}
+                    // [ Hash(Hash_t|Hash_partial_priduct) ]
+                    // [ Hash(G2Acc) ]
+                }
+            })
+            .clone()
```
storopoli, line 165: Why do we need to `clone` inside the cache?

bitvm/src/chunk/wrap_hasher.rs
```diff
@@ -41,113 +42,148 @@ pub(crate) mod hash_utils {
...
+                    {hash_fp14()}
+                    {Fq::fromaltstack()}
+                    {Fq::roll(1)}
+                    // [ Hash_t, Hash_partial_product ]
+                    {hash_fp2()}
+                    // [ Hash(Hash_t|Hash_partial_priduct) ]
+                    // [ Hash(G2Acc) ]
+                }
+            })
+            .clone()
```
storopoli, line 186: Why do we need to `clone` inside the cache?

storopoli (changes requested):
Can't all these `clone`s be statics refs to `&Script`?

bitvm/src/chunk/wrap_hasher.rs
sander2: Beause `OnceLock::get_or_init` is returning `&T`

sander2:
> Can't all these `clone`s be statics refs to `&Script`?

In theory the functions could return `&Script`, yes. However, this PR is just for some low hanging fruit - it reduces execution time from 270s to 90s. It might be possible to shave off a couple of milliseconds with your suggestion but I don't think that's worth it. The change would break the `script!` macro, meaning either we'd need a change in the rust-bitcoin-script library to allow pushes of `&Script` rather than `Script`, or we would need to clone at the calling site

bitvm/src/chunk/wrap_hasher.rs
storopoli: Why not `&'static T`?

AaronFeickert (approved):
This looks fine. I think you could simplify the construction somewhat using `LazyLock`, but that's really just a nit.

storopoli (approved):
ACK ff73707b7aafd7c5fb93b78a80b3e2c7902e77c4

Agree with @AaronFeickert's `LazyLock` commentc

AaronFeickert:
Just realized that this drops the `test_script` CI runtime from 43 minutes down to 24 minutes! Thanks for the contribution :D

AaronFeickert (approved):
ACK 8ef4f0b

PR #297, opened 2025-05-30
Fix/disprove script exceeds 4 m with constraints

### LINKED ISSUE #296: The largest Disprove tx too large to be included in a block under default Bitcoin Core settings
Hi there,

We, the Babylon team, have encountered a practical issue with the largest *Disprove* tx size that may impact the usability of the BitVM bridge design in real-world deployments.

### Context

- The *Assert* transaction in our setup contains **378 assertions**.
- Through analysis, we found that the **largest Disprove tx** corresponds to **assertion index 372**.
- We're simulating a case where this assertion is invalid and must be disproved on-chain.

### The Problem

The resulting Disprove tx is **3,996,294 weight units**, but cannot be mined under **default Bitcoin Core settings** due to two key constraints:

#### Before Bitcoin Core 29.0

- Default `blockmaxweight` is `MAX_BLOCK_WEIGHT - 4000 = 3,996,000`, which is already too small for this Disprove tx.
    - Ref: https://github.com/bitcoin/bitcoin/blob/110183746150428e6385880c79f8c5733b1361ba/src/policy/policy.h#L23C56-L23C79
- Many miners are still on versions < 29.0, so this limitation is highly relevant today.

#### Starting from Bitcoin Core 29.0

- `blockmaxweight` now defaults to `MAX_BLOCK_WEIGHT`, which looks promising.
    - Ref: https://github.com/bitcoin/bitcoin/blob/f490f5562d4b20857ef8d042c050763795fd43da/src/policy/policy.h#L23
- However, a new flag `blockreservedweight=8000` is introduced, reserving 8000 weight units for fixed block overhead (header, tx count, coinbase).
    - Ref: https://github.com/bitcoin/bitcoin/blob/f490f5562d4b20857ef8d042c050763795fd43da/src/policy/policy.h#L25
- This further reduces the space available for normal transactions, again making the Disprove tx inclusion impossible in reality.

### Impact

Unless miners override defaults, **this tx will not be accepted into blocks,** making BitVM bridges unusable in worst-case assertion scenarios.

manishbista28:
Fixes #296 

Problem: Disprove script of max size did not have enough space for `blockreservedweight`

Max script size is now reduced by ~48K opcodes.
This should be sufficient to account for block `blockreservedweight ` of around 8000.

The size of max disprove script is around 3953286 opcodes with max runtime stack use of 933.

Changes:
I added a different version of tmul for Fp2 multiplication.
`fp_lc_mul!(Mul2LCW4, 4, 4, [true, true]);`
 This uses more stack size but lesser script.

Added code inside fp254impl.rs, fq.rs and fq2.rs are direct copies of their window=3 counterparts with change only on the called function "tmul_lc2_w4".

I have used this version of multiplication in the script corresponding to largest disprove script.
Finally I executed test_fn_disprove_invalid_assertions to verify that Script and stack use are within limits for the tapscripts.

### DIFF
--- a/bitvm/src/bn254/fp254impl.rs
+++ b/bitvm/src/bn254/fp254impl.rs
@@ -699,7 +699,7 @@ pub trait Fp254Impl {
     ) -> (Script, Vec<Hint>) {
         assert!(a_depth > b_depth && b_depth > c_depth && c_depth > d_depth);
 
-        let mut hints = Vec::new();
+        let mut hints = Vec::with_capacity(1);
 
         let modulus = &Fq::modulus_as_bigint();
 
@@ -845,7 +845,7 @@ pub trait Fp254Impl {
     ) -> (Script, Vec<Hint>) {
         assert!(a_depth > b_depth && b_depth > c_depth && c_depth > d_depth);
 
-        let mut hints = Vec::new();
+        let mut hints = Vec::with_capacity(1);
 
         let modulus = &Fq::modulus_as_bigint();
 
--- a/bitvm/src/bn254/fp254impl.rs
+++ b/bitvm/src/bn254/fp254impl.rs
@@ -686,6 +686,9 @@ pub trait Fp254Impl {
         (script, hints)
     }
 
+    // Assumes tmul hint (1 BigInteger) at the top of stack
+    // and operands (a, b, c, d) at stack depths (a_depth, b_depth, c_depth, d_depth)
+    // Computes r = a * c + b * d (mod p)
     #[allow(clippy::too_many_arguments)]
     fn hinted_mul_lc2_w4(
         a_depth: u32,
@@ -832,6 +835,7 @@ pub trait Fp254Impl {
         (script, hints)
     }
 
+    // Same as hinted_mul_lc2_keep_elements(), except retains operands (a, b, c, d) on stack
     #[allow(clippy::too_many_arguments)]
     fn hinted_mul_lc2_keep_elements_w4(
         a_depth: u32,
--- a/bitvm/src/bn254/fp254impl.rs
+++ b/bitvm/src/bn254/fp254impl.rs
@@ -687,6 +687,46 @@ pub trait Fp254Impl {
     }
 
     #[allow(clippy::too_many_arguments)]
+    fn hinted_mul_lc2_w4(
+        a_depth: u32,
+        a: ark_bn254::Fq,
+        b_depth: u32,
+        b: ark_bn254::Fq,
+        c_depth: u32,
+        c: ark_bn254::Fq,
+        d_depth: u32,
+        d: ark_bn254::Fq,
+    ) -> (Script, Vec<Hint>) {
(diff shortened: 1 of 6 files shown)


manishbista28:
Disprove Script with max Script size: 
7-th last index
3953286 opcodes @ 933 max runtime stack

Disprove Script with max stack use: 
6-th last index
3471804 opcodes @946 max runtime stack.

bitvm/src/bn254/fp254impl.rs
```diff
@@ -686,6 +686,46 @@ pub trait Fp254Impl {
...
+        b_depth: u32,
+        b: ark_bn254::Fq,
+        c_depth: u32,
+        c: ark_bn254::Fq,
+        d_depth: u32,
+        d: ark_bn254::Fq,
+    ) -> (Script, Vec<Hint>) {
+        assert!(a_depth > b_depth && b_depth > c_depth && c_depth > d_depth);
+
+        let mut hints = Vec::new();
```
liuchengxu, line 702: Looks like the size is known beforehand, can we make it explicit? Same to the others.

```suggestion
        let mut hints = Vec::new_capacity(1);
```

manishbista28: fixed in latest commit. 
I have made the size specification only for changes scoped within this PR.

bitvm/src/bn254/fp254impl.rs
```diff
@@ -686,6 +686,46 @@ pub trait Fp254Impl {
         (script, hints)
     }
 
+    #[allow(clippy::too_many_arguments)]
+    fn hinted_mul_lc2_w4(
```
AaronFeickert, line 690: Please document.

bitvm/src/bn254/fp254impl.rs
```diff
@@ -792,6 +832,46 @@ pub trait Fp254Impl {
         (script, hints)
     }
 
+    #[allow(clippy::too_many_arguments)]
+    fn hinted_mul_lc2_keep_elements_w4(
```
AaronFeickert, line 836: Please document.

bitvm/src/bn254/fq.rs
```diff
@@ -47,6 +47,12 @@ impl Fq {
         }
     }
 
+    pub fn tmul_lc2_w4() -> Script {
```
AaronFeickert, line 50: Please document.

bitvm/src/bn254/fq2.rs
```diff
@@ -192,6 +192,42 @@ impl Fq2 {
         (script, hints)
     }
 
+    pub fn hinted_mul_w4(
```
AaronFeickert, line 195: Please document.

AaronFeickert (changes requested):

wz14 (commented):
Thanks! LGTM

bitvm/src/bn254/fp254impl.rs
manishbista28: added

PR #299, opened 2025-06-19
Add error handling for Musig2 signature generation in pre_signed_musig2.rs

GarmashAlex:
Replaced all usages of .unwrap() with proper error handling in the pre_sign_musig2_taproot_input and finalize_musig2_taproot_input functions.
Now, if signature generation fails, an error message is printed and the function returns early instead of panicking.
This improves the robustness of Musig2 transaction signing and prevents unexpected crashes during transaction processing.
No changes were made to function signatures; errors are logged and execution is safely aborted in case of failure.

### DIFF
--- a/bridge/src/transactions/pre_signed_musig2.rs
+++ b/bridge/src/transactions/pre_signed_musig2.rs
@@ -124,7 +124,7 @@ pub fn pre_sign_musig2_taproot_input<T: PreSignedTransaction + PreSignedMusig2Tr
     let script = &tx.prev_scripts()[input_index].clone();
     let musig2_nonces = &tx.musig2_nonces()[&input_index].values().cloned().collect();
 
-    let partial_signature = generate_taproot_partial_signature(
+    let partial_signature = match generate_taproot_partial_signature(
         context,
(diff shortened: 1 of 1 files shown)


Gb2real (approved):

bridge/src/transactions/pre_signed_musig2.rs
```diff
@@ -124,7 +124,7 @@ pub fn pre_sign_musig2_taproot_input<T: PreSignedTransaction + PreSignedMusig2Tr
     let script = &tx.prev_scripts()[input_index].clone();
     let musig2_nonces = &tx.musig2_nonces()[&input_index].values().cloned().collect();
 
-    let partial_signature = generate_taproot_partial_signature(
+    let partial_signature = match generate_taproot_partial_signature(
```
eigmax, line 127: if we return before updating the witness, there is no way to catch such exceptions. 

Besides, there is an `expect` in `generate_taproot_partial_signature`,  my suggestion would be panic immediately and throw the reason. 

how do you think, @AaronFeickert ?

PR #300, opened 2025-07-17
Clear Part1 Report

Hakkush-07:
3.1: BigIntImpl::limb_from_bytes and BigIntImpl::from_bytes functions are simply unused. I removed them.
3.2: limb_to_le_bits_toaltstack function considers the case when n_bits=1 and n_bits=0, suggested solution.
3.3, 3.5: double_allow_overflow functions now expects the reference in BigInt units as in other functions, also hardcoded "9" is fixed.
3.4: I added an assertion to BigIntImpl struct that makes sure N_LIMBS is at least 2.
3.6: make sure most significant 1 lies inside the limits.
3.7: incorrect powers of two for n<7, although we dont encounter this case, is fixed by applying the suggested.
3.8: Bigint mul is simply unused. I removed it.
3.9: applied the suıggested solution.
3.10: u4_add_tables optimization is implemented and necessary hardcoded table size counts are updated.


bitvm/src/bigint/inv.rs
```diff
@@ -48,7 +48,7 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
 pub fn limb_shr1_carry(num_bits: u32) -> Script {
     let powers_of_2_script = if num_bits < 7 {
         script! {
-            for i in 0..num_bits - 1 {
+            for i in 1..num_bits {
```
eigmax, line 51: Can you please update the test `test_limb_shr1_carry` to run on different num_bits, now 29 is hardcode, we can try some different numbers which maybe less than 7.

bitvm/src/lib.rs
```diff
@@ -166,6 +166,16 @@ fn execute_script_buf_optional_stack_limit(
             break;
         }
     }
+    for (i, a) in exec.stack().iter_str().enumerate() {
```
eigmax: Should we remove this?

eigmax (approved):

Issue #301, opened 2025-07-24
[Zellic Audit] Hash Bug


wz14:
> (Mark 3 09.07) in Hakan's bug list

Here are some related messages in the telegram group.

Finding: Hash validation of inputs can erroneously succeed on ill-formed input

For some U256-values x, it is possible to construct bitcoin stack values y of the same stack size (9 bitcoin ints) which do not represent a valid U256, but are accepted by the hash function and produce the same hash as x.

Example: For x = 0 (limbs: [0,0,0,0,0,0,0,0]), and the value y=[0,0,0,0,0,0,0,0,-1], the pairs (x,x) and (x,y) produce the same hash when blake_compute_script(64).

This in allows the following attack: The disprover provides ill-formed input hints that have the same hash as the asserted inputs.
The ill-formed input hints produce different well-formed results after applying the op_script (this is plausible, e.g. after adding 1 to the ill-formed y above, the result is a wellformed 0, which differs from adding 1 to x, which is 1).
Then the hash-script erroneously confirms that the ill-formed input was the asserted one (as it hashes to the correct value). but the output differs, so the disprover succeeds in claiming the reward even though the proof was valid.

A hash-collision of this form can be found if the correct value contains a limb that is either 0 or the max allowed value for that limb. This can likely occur for example at the beginning of the miller loop, where the accumulator contains 0-limbs

```rust
    #[test]
    fn test_zellic_hash_collision_on_invalid_input() {
        let zero = script! {
            {0} {0} {0} {0} {0} {0} {0} {0} {0}
        };
        let fake_zero = script! {
            {0} {0} {0} {0} {0} {0} {0} {0} {-1}
        };
        let res = execute_script(script! {
            // fake_zero has same hash as zero
            {zero.clone()} {zero.clone()} {blake3_compute_script(64)}
            for _ in 0..64 {
                OP_TOALTSTACK
            }
            {zero.clone()} {fake_zero.clone()} {blake3_compute_script(64)}
            for i in 0..64 {
                OP_FROMALTSTACK
                {64-i} OP_ROLL
                OP_EQUALVERIFY
            }
            // fake_zero + (1: Fq) = (0: Fq)
            {Fq::push(1.into())} {fake_zero.clone()} {Fq::add(1, 0)}
            {Fq::push(0.into())} {Fq::equalverify(1, 0)}
            {1}
        });
        println!{"{:?} {:?}", res.success, res.final_stack};
    }
```

Hakkush-07:
[...]

PR #316, opened 2025-07-25
[Zellic Audit] Hash Bug (Mark 3 09.07)

wz14:
This PR targets https://github.com/BitVM/BitVM/issues/301

We convert a 29-limb bigint into a 4-limb by function `transform_limbsize` in the Blake3. The `transform_limbsize` extracts digits for the target limb by comparing the source limb with a 2-power positive integer. If there is a {-1} limb in an integer, then the extracted digits are all zeros, which is the same as the limb {0}. 

```
        let fake_zero = script! {
            {0} {0} {0} {0} {0} {0} {0} {0} {-1}
        };
        let zero = script! {
            {0} {0} {0} {0} {0} {0} {0} {0} {0}
        };
```

It turns out that there is no valid check when we try to use `transform_limbsize`. I add a function `verify_bigint_on_stack` to check if the data on the stack is a valid bigint. Now, the script of the Blake3 function will reject the invalid input by failing the script.

### DIFF
--- a/bitvm/src/bigint/std.rs
+++ b/bitvm/src/bigint/std.rs
@@ -488,6 +488,81 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
             )
         }
     }
+
+    /// Validate that the BigInt on stack has valid limb values
+    ///
+    /// This function checks that each limb in the BigInt representation
+    /// does not exceed the maximum value allowed for the given LIMB_SIZE
+    /// and ensures no limb is negative.
+    ///
+    /// ## Stack Effects:
+    /// - Input: BigInt limbs (MSB first, LSB on top)
+    /// - Output: Same BigInt limbs + validation result (1 if valid, 0 if invalid)
+    ///
+    /// ## Validation Rules:
+    /// - Each limb must be >= 0 (no negative values)
+    /// - Each limb must be < (1 << LIMB_SIZE)
+    /// - The head limb must be < (1 << HEAD_OFFSET) where HEAD_OFFSET is the remaining bits
+    ///
+    /// ## Note:
+    /// This function is expensive in terms of script size and should be used carefully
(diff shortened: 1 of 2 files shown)


bitvm/src/hash/blake3.rs
```diff
@@ -607,4 +611,34 @@ mod tests {
     fn test_maximum_alstack_element_calculation_with_all_limbs() {
         test_maximum_alstack_element_calculation_with_limbs(&ALL_POSSIBLE_LIMB_LENGTHS);
     }
+
+    #[test]
+    fn test_hash_collision_on_invalid_input() {
```
sander2: Small nit: with the new code this doesn't actually test for a hash collision - the script will fail inside the `blake3_compute_script` on the `fake_zero`, so the rest could technically be removed. So it's more of a `test_failure_on_invalid_input`

bitvm/src/bigint/std.rs
```diff
@@ -488,6 +488,81 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
...
+    /// - Input: BigInt limbs (MSB first, LSB on top)
+    /// - Output: Same BigInt limbs + validation result (1 if valid, 0 if invalid)
+    ///
+    /// ## Validation Rules:
+    /// - Each limb must be >= 0 (no negative values)
+    /// - Each limb must be < (1 << LIMB_SIZE)
+    /// - The head limb must be < (1 << HEAD_OFFSET) where HEAD_OFFSET is the remaining bits
+    ///
+    /// ## Note:
+    /// This function is expensive in terms of script size and should be used carefully
```
sander2, line 508: we'll need to remember to reevaluate that the largest chunk still fits into a single tx, after all the changes are done

sander2 (commented):
lgtm

bitvm/src/hash/blake3.rs
wz14: Changed the name and remove unreachable scripts.

PR #321, opened 2025-08-04
fix: correct u8_add modulo 256 calculation

MozirDmitriy:
Fix logical error in u8_add() function where OP_0 was incorrectly used 
instead of proper modulo 256 calculation.

The function now correctly implements (a + b) % 256 by:
- Adding two u8 values
- Checking if result >= 256
- Subtracting 256 when overflow occurs
- Returning the modulo result

This ensures proper u32 addition behavior when handling byte-level 
arithmetic operations.


sander2 (changes requested):
I don't think this change is correct. I think the op_0 is needed because of the op_drop below. If you remove it, the op_drop would drop the output value

PR #328, opened 2025-08-06
[Zellic Audit] fix fp254impl issues

Hakkush-07:
This PR fixes findings (Malte 1), (Malte 15), (Mohit 12) in the list.

### Malte 1
Optimization: `ROT, SWAP, ROT` in `Fp254Impl::add`

In Fp254Impl::add in bitvm/src/bn254/fp254impl.rs, there is the following passage:
rust   
                    OP_ROT
                    OP_SWAP
                    OP_ROT
this is equivalent to a single OP_SWAP.

### Malte 15
Observation: Functions specific for `Fq` implemented in the `Fp254Impl` trait

Various functions are implemented for the Fp254Impl trait that however operate specifically on Fq, which is only one specific type that implements Fp254Impl. The type Fr is another one implementing Fp254Impl, so counterintuitively, Fr::hinted_mul is a function operating on Fq rather than Fr.

Note also that some of these functions utilize Self, which might refer to an uninted type, for example Fr, instead of the intended Fq. At the moment, the usages of Self are only related to accessing the number of limbs and similar, which are the same in all implementations of Fp254Impl present in the current code base.

These issues appear in the functions in bitvm/src/bn254/fp254impl.rs from hinted_mul to hinted_inv.

While we have not observed an issue in the current state of the codebase connected to this (so far at least), unintuitive behavior like this can cause developer mistakes in the future.

We recommend to move these functions to an implementation for Fq specifically.

### Mohit 12
Fp254Impl::neg has a manually unrolled loop and will fail if N_LIMBS is not 9.

### DIFF
--- a/bitvm/src/bn254/fp254impl.rs
+++ b/bitvm/src/bn254/fp254impl.rs
@@ -1,14 +1,11 @@
 use crate::bigint::add::limb_add_carry;
 use crate::bigint::sub::limb_sub_borrow;
 use crate::bigint::U254;
-use crate::bn254::fq::Fq;
-use crate::bn254::utils::Hint;
 use crate::treepp::*;
 use ark_ff::PrimeField;
 use bitcoin_script::script;
-use num_bigint::{BigInt, BigUint};
+use num_bigint::BigUint;
 use num_traits::Num;
-use std::str::FromStr;
 use std::sync::OnceLock;
 
 #[allow(clippy::declare_interior_mutable_const)]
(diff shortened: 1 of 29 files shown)


eigmax (approved):

bitvm/src/bn254/fp254impl.rs
```diff
@@ -275,51 +270,25 @@ pub trait Fp254Impl {
...
-                OP_NIP OP_ADD
-                // ⋯ C₇⁻+A₈
-                { Self::MODULUS_LIMBS[8] } OP_SWAP OP_SUB
-                // ⋯ M₈-(C₇⁻+A₈)
-                OP_FROMALTSTACK OP_FROMALTSTACK OP_FROMALTSTACK OP_FROMALTSTACK
-                OP_FROMALTSTACK OP_FROMALTSTACK OP_FROMALTSTACK OP_FROMALTSTACK
+                for i in 0..Self::N_LIMBS-1 {
+                    { Self::MODULUS_LIMBS[i as usize] } OP_SWAP
+                    if i == 0 {
+                        { 2_usize.pow(U254::LIMB_SIZE) }
```
eigmax, line 269: One question here:  we can use 0x20000000 here for whatever N_LIMBS, yes?

PR #329, opened 2025-08-08
fix #302 G1Affine::hinted_check_add has a soundness issue

### LINKED ISSUE #302: [Zellic Audit] G1Affine::hinted_check_add has a soundness issue (Mohit 2 11.06)

lynndell2010:
fix #302 
// add a judge: t and q are not equal. If they are equal, then call hinted_check_double().
if !t.is_zero() && !q.is_zero() && (t.x != q.x)
This judgment should also be added to the script.

### DIFF
--- a/bitvm/src/bn254/g1.rs
+++ b/bitvm/src/bn254/g1.rs
@@ -4,6 +4,7 @@ use crate::bn254::utils::Hint;
 use crate::treepp::{script, Script};
 use ark_ec::AffineRepr;
 use ark_ff::{AdditiveGroup, Field};
+use bitcoin::opcodes::all::OP_ELSE;
 use num_bigint::BigUint;
 
 use super::fq2::Fq2;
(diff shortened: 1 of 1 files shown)


sevkett12:
- We can't use G1::equalverify() because it doesn't return 0/1 instead it just verifies which we do not want. If you meant to use G1::equal() (which doesn't exist and should be created), in case of equality  we should use G1::double() but you use is_zero_keep_element (which seemed irrelevant to me). And also the test bn254::g1::test::test_g1_affine_hinted_check_add doesn't work. However this makes the script a lot bigger.
-  I think since we are using G1::hinted_check_add is used only for scalar_mul, it can be mathematically proven that the points we give to this function cannot be equal
- I can also suggest that we use two different functions for G1 addition one of which assumes the caller ensures the points are different while the other doesn't but we should continue to use the existing one.

PR #331, opened 2025-08-08
[Zellic Audit] tmul: check hints are valid bigints

### LINKED ISSUE #304: [Zellic Audit] tmul: not checking for the limbs (Malte 5 16.06)
I don't have an end-to-end proof of concept for tmul, but here is an example showing that not checking for the limbs being in the expected range breaks fundamental assumptions:

The limb_add_carry function in bitvm/src/bigint/add.rs gets as input two limbs a and b, and the base (say 2^29), and is intended to return the base, the carry, and (a+b) % base. One fundamental property of addition is that it is assocative. But the implementation of limb_add_carry fails to be associative when allowing non-standard inputs.

The way it is implemented is that it computes a+b, and then if a+b >= base it sets carry to 1 and subtracts base from a+b, and otherwise carry is set to 0.

Now suppose we have three limbs to add: -2^29, 2^28 and 2^28, with base=2^29.

If we first use limb_add_carry on the first two, then we will get as result -2^28, with no carry. Applying limb_add_carry on this with the last limb, we then get as result 0 with still no carry.

However, if we group the two 2^28 together first, that results in a carry with intermediate result 0, and then the final result will be -2^29 with one carry overall.

So if the hints being well-formed values is not checked, then reasoning about the correctness of for example tmul can not use the property of the bigint arithmetic of being associative, and similarly various other reasonable properties.

Note that if one reconstructs the value from the limbs as sum_i limb_i * 2^{29*i}, then one can consider (1, -2^29) a non-standard variant of (0, 0) representing the same value. So one might think that means this isn't a problem, it is just a non-standard representation. But for example the BigIntImpl::equal function will not recognize them as equal and so will give an incorrect result. More relevant is probably BigIntImpl::lessthan: I haven't gone through this one myself, but I strongly suspect that it will also result in an incorrect result in which (1, -2^29) is considered bigger than (0, a) for any a. So the upshot is that at the latest when one starts using lessthan and then branch on the result of that, things can go differently than expected if having used not-validated bigints initially caused an input to be such a non-standard value. It then becomes very hard to reason about the scripts.

I think it is quite plausible that there could be a way to leverage such misbehavior of basic bigint functions on invalid inputs to prove an incorrect result for tmul. For example, at the end lessthan is used in a check like x*y - q*p < p (unsigned comparison). If one got such an invalid bigint as x*y - q*p because q was invalid, then one might pass this check even though the result is an invalid value that also would not represent an integer in the intended range.

### LINKED ISSUE #305: [Zellic Audit] tmul: accept non-standard inputs without breaking (Malte 7 18.06)
followup of Malte 5

The example I give with tmul followed by div3 shows that option 2 would not be enough, you would need to upgrade a large number of functions (such as div3) to also accept non-standard inputs without breaking.

The div3 script does not just transform a non-standard input to a non-standard output that is otherwise correct, it produces an actually fully incorrect result

### LINKED ISSUE #307: [Zellic Audit] tmul: Incorrect results of `var_windows_script` (Malte 8 25.06)
One important part of the tmul implementation in bitvm/src/bn254/fq.rs is the var_windows_script sub-script, which is used to extract windows from the factors y. The main loop that extracts and accumulates the relevant bits is the following:

```
                                for i in st..VAR_WIDTH {
                                    if s_limb > e_limb {
                                        if i % LIMB_SIZE == (s_bit % LIMB_SIZE) + 1 {
                                            // window is split between multiple limbs
                                            { stack_top + T::N_LIMBS * j + e_limb + 1 } OP_PICK
                                        }
                                    }
                                    if ( i == 0){
                                        { 1 << ((s_bit - i) % LIMB_SIZE) }
                                        OP_2DUP
                                        OP_GREATERTHANOREQUAL
                                        OP_IF
                                            OP_SUB
                                            2
                                        OP_ELSE
                                            OP_DROP
                                            0
                                        OP_ENDIF
                                        OP_SWAP
                                    } else{
                                        if (s_bit - i) % LIMB_SIZE > 7 {
                                            { 1 << ((s_bit - i) % LIMB_SIZE) }
                                            OP_2DUP
                                            OP_GREATERTHANOREQUAL
                                            OP_IF
                                                OP_SUB
                                                OP_SWAP OP_1ADD
                                            OP_ELSE
                                                OP_DROP
                                                OP_SWAP
                                            OP_ENDIF
                                            if i < VAR_WIDTH - 1 { { NMUL(2) } }
                                            OP_SWAP
                                        } else {
                                            OP_TUCK
                                            { (1 << ((s_bit - i) % LIMB_SIZE)) - 1 }
                                            OP_GREATERTHAN
                                            OP_TUCK
                                            OP_ADD
                                            if i < VAR_WIDTH - 1 { { NMUL(2) } }
                                            OP_ROT OP_ROT
                                            OP_IF
                                                { 1 << ((s_bit - i) % LIMB_SIZE) }
                                                OP_SUB
                                            OP_ENDIF
                                        }
                                    }
                                }
```

### LINKED ISSUE #310: [Zellic Audit] The `init_table` script in `tmul` does not check for overflow (Malte 10 01.07)
This finding is about the init_table function used in tmul, in particular its application to -q. The purpose of the function is to place the values -2*q, ..., -(2^w - 1)*q on the stack (where w is the window size used).
It is important that overflows are prevented: If a signed overflow would happen, the script should fail instead, as discussed for previous findings.
However, init_table does not ensure this. It is implemented as follows:
```
fn init_table(window: u32) -> Script {
    assert!(
        (1..=6).contains(&window),
        "expected 1<=window<=6; got window={}",
        window
    );
    script! {
        for i in 2..=window {
            for j in 1 << (i - 1)..1 << i {
                if j % 2 == 0 {
                    { T::double_allow_overflow_keep_element( (j/2 - 1) * T::N_LIMBS ) }
                } else {
                    { T::add_ref_with_top(j - 2) }
                }
            }
        }
    }
}
```
As the name suggests, double_allow_overflow_keep_element actually does allow overflows. It is thus possible for certain values of q to obtain tables containing incorrect values due to overflows, without the script failing.

### LINKED ISSUE #311: [Zellic Audit] The `tmul` script is unable to handle small negative results (Malte 14 02.07)
The fp_lc_mul macro in bitvm/src/bn254/fq.rs can be used to implement different variants of multiplication for the Fq field, via a function tmul. The inputs to tmul are values x_i and y_i that are bigints representing elements of the field Fq, and tmul is inteded to compute sum_{i=0}^{L-1} s_i * x_i * y_i with arithmetic in the field Fq, with s_i fixed constant signs, that is s_i in {-1, 1}.

To help with this, tmul also receives a hint q for the quotient of the division of sum_{i=0}^{L-1} s_i * x_i * y_i, computed with integer arithmetic, by the prime order p of Fq. There are different rounding conventions one can take, and the Rust bigint type used to handle hints in the Rust code round towards zero.

Hakkush-07:
This PR adds check_validity() function BigIntImpl which checks every limb is in correct interval. This function is then used in tmul to check the given hint is a well-formed bigint. This adds +70 bytes of script to Fq::hinted_mul (67820 -> 67890). 

To reduce this, I also noticed a small optimization while fixing this. It is to replace `T::push_zero() + T::sub()` with `T::neg()`. neg function doesnt exist, so I created it, it basically first performs `(2^N_BITS-1)-x`, and then adds 1. This takes Fq::hinted_mul from 67890 to 67837. 

Closes #304 and closes #305 (followup). (edit: also closes #307 ) (edit: also closes #311 ) (edit: also closes #310 )

### DIFF
--- a/bitvm/src/bigint/std.rs
+++ b/bitvm/src/bigint/std.rs
@@ -307,6 +307,24 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
         }
     }
 
+    /// checks if the element on top of the stack is a valid bigint, needed for hints
+    /// - head limb in the range [0..HEAD_OFFSET-1]
+    /// - other limbs in the range [0..2^LIMB_SIZE-1]
+    pub fn check_validity() -> Script {
+        script! {                            // a0 a1 ... an
+            { 1 << LIMB_SIZE }               // a0 a1 ... an x
+            for _ in 0..Self::N_LIMBS-2 {    // a x
+                OP_TUCK                      // x a x
+                0 OP_SWAP                    // x a 0 x
+                OP_WITHIN OP_VERIFY          // x
+            }                                // a0 a1 x
+            0 OP_SWAP                        // a0 a1 0 x
+            OP_WITHIN OP_VERIFY              // a0
+            0 { Self::HEAD_OFFSET }          // a0 0 y
+            OP_WITHIN OP_VERIFY
+        }
+    }
+
     /// Resize positive numbers
     ///
     /// # Note
--- a/bitvm/src/bigint/add.rs
+++ b/bitvm/src/bigint/add.rs
@@ -63,6 +63,34 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
         }
     }
 
+    /// add one
+    pub fn add1() -> Script {
+        script! {
+            OP_1ADD                             // a0 ... an
+            { 1 << LIMB_SIZE }                  // a0 ... an x
+            OP_SWAP                             // a0 ... x an
+            for _ in 0..Self::N_LIMBS-1 {       // ... x a
+                OP_2DUP                         // ... x a x a
+                OP_EQUAL                        // ... x a 0/1
+                OP_TUCK                         // ... x 0/1 a 0/1
+                OP_IF OP_NOT OP_ENDIF           // ... x 0/1 a/0
+                OP_TOALTSTACK                   // ... x 0/1
+                OP_ROT                          // .. x 0/1 a_{i-1}
+                OP_ADD                          // .. x a'
+            }
+                                                // x a0
+            OP_NIP                              // a0
+            { Self::HEAD_OFFSET }               // a0 y
+            OP_OVER                             // a0 y a0
+            OP_EQUAL                            // a0 0/1
+            OP_IF OP_NOT OP_ENDIF               // a/0
+
+            for _ in 0..Self::N_LIMBS - 1 {
+                OP_FROMALTSTACK
+            }
+        }
+    }
+
     /// Double the BigInt on top of the stack
     ///
     /// # Note
@@ -531,6 +559,25 @@ mod test {
                 { U64::add(1, 0) }
                 { U64::push_u64_le(&[c]) }
                 { U64::equalverify(1, 0) }
+                OP_TRUE
+            };
+            run(script);
+        }
+    }
+
+    #[test]
+    fn test_add1() {
+        println!("U254.add1: {} bytes", U254::add1().len());
+        let mut prng = ChaCha20Rng::seed_from_u64(0);
+        for _ in 0..100 {
+            let a: BigUint = prng.sample(RandomBits::new(254));
+            let c: BigUint = (a.clone() + BigUint::one()).rem(BigUint::one().shl(254));
+
+            let script = script! {
+                { U254::push_u32_le(&a.to_u32_digits()) }
+                { U254::add1() }
+                { U254::push_u32_le(&c.to_u32_digits()) }
+                { U254::equalverify(1, 0) }
                 OP_TRUE
             };
             run(script);
--- a/bitvm/src/bigint/add.rs
+++ b/bitvm/src/bigint/add.rs
@@ -172,6 +172,32 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
         }
     }
 
+    pub fn double_prevent_overflow_keep_element(n: u32) -> Script {
+        script! {
+            for _ in 0..Self::N_LIMBS {
+                { n + Self::N_LIMBS - 1 } OP_PICK
+            }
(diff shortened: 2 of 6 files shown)


Hakkush-07:
Also, fixes issue when LCS is not power of two, and asserts window is not 1. Closes #307 (Malte 8 and Malte 9)

sevkett12:
I removed the positivity check from tmul. Caller now must ensure the hint is compatible. Fq::hinted_mul is now  
67837 -> 67604

Hakkush-07:
Added code clarity suggestions (Malte 11, Malte 13, Malte 16, Malte 18)

sevkett12:
I changed T::check_validity to U::check_validity because q needs to be a valid U BigInt. Additionally, I added double_keep_element_prevent_overflow and test the tmul with it. It works fine because we check q beforehand but it is less efficient and not needed as long as q is U BigInt.  With that we don't need to do additional fix to issue #310

Hakkush-07:
tmul calculates $c=\left(\sum_{i=1}^L s_ix_iy_i\right)-pq$ such that $0\leq c\leq p-1$ where $x_i,y_i\in Fq$ and $s_i\in\{-1,1\}$. We use U to represent Fq elements which includes $x_i$ and $y_i$. when $L=1$, $q$ also needs as many bits as a Fq has. But for $L\geq 2$, we have $|q|< \frac{Lp^2}{p}=Lp$, and $Lp$ has `Fq::N_BITS+log2(L)` bits. We also use T type to represent the running accumulator, which might have one more bit than $2^w q$ has. We need to check validity of the hint q, that is, it fits into the correct number of bits, and that is neither U nor T. For LC=1 normal modular multiplication, U works, but we need a third type in general, and I defined it as W.

bitvm/src/bigint/std.rs
```diff
@@ -307,6 +307,24 @@ impl<const N_BITS: u32, const LIMB_SIZE: u32> BigIntImpl<N_BITS, LIMB_SIZE> {
         }
     }
 
+    /// checks if the element on top of the stack is a valid bigint, needed for hints
+    /// - head limb in the range [0..HEAD_OFFSET-1]
+    /// - other limbs in the range [0..2^LIMB_SIZE-1]
+    pub fn check_validity() -> Script {
```
eigmax, line 313: This is duplicated with #316, but fix different functions.

Hakkush-07:
other PRs take place of this big one

PR #334, opened 2025-08-11
Zellic Audit G1 fixes

sevkett12:
I fixed Mohit 3,4,6,8.
For Mohit 3 : I added a check for infinity case. Additionally G2Affine::read_from_stack had the same problem and I fixed that too. 
For Mohit 4: I deleted the function G1Affine::zero() and replaced it with G1Affine::identity()
For Mohit 6: We implemented the optimization and makethe same optimization for G2Affine::hinted_is_on_curve  
For Mohit 8 : Since the functions hinted_x_from_eval_point and hinted_y_from_eval_point are always being used together we delete them and I wrote the functions inside the hinted_from_eval_point and decreased the number of assertions to 1. 

**Mohit 3**
G1Affine::read_from_stack sets infinity to false even if the point being read is the zero point.

**Mohit 4**
G1Affine::push_zero does the same thing as G1Affine::identity and is not used anywhere except dfs_with_constant_mul in msm so can be removed

**Mohit 6**
The Fq::roll(1) in G1Affine::hinted_is_on_curve can also be avoided if y_sq is executed as the first instruction in the script

**Mohit 8**
Optimization: In bn254/[g1.rs](https://g1.rs/), in both hinted_from_eval_points and hinted_from_eval_point,  yinv*y == 1 is checked twice because it's checked in both hinted_x_from_eval_point and hinted_y_from_eval_point. The check can probably be removed from one of these 2 to save opcodes in hinted_from_eval_points

### DIFF
--- a/bitvm/src/bn254/g1.rs
+++ b/bitvm/src/bn254/g1.rs
@@ -147,7 +147,7 @@ impl G1Affine {
             witness[Fq::N_LIMBS as usize..2 * Fq::N_LIMBS as usize].to_vec(),
         ))
         .into();
-        let is_inf = x.is_zero() & y.is_zero();
+        let is_inf = x.is_zero() && y.is_zero();
         ark_bn254::G1Affine {
             x,
             y,
--- a/bitvm/src/bn254/g1.rs
+++ b/bitvm/src/bn254/g1.rs
@@ -132,13 +132,6 @@ impl G1Affine {
         (script, hints)
     }
 
-    pub fn push_zero() -> Script {
-        script! {
-            { Fq::push_zero() }
-            { Fq::push_zero() }
-        }
-    }
-
     pub fn push(element: ark_bn254::G1Affine) -> Script {
         script! {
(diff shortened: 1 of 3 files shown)


bitvm/src/bn254/g1.rs
```diff
@@ -150,10 +143,15 @@ impl G1Affine {
         assert_eq!(witness.len() as u32, Fq::N_LIMBS * 2);
         let x = Fq::read_u32_le(witness[0..Fq::N_LIMBS as usize].to_vec());
         let y = Fq::read_u32_le(witness[Fq::N_LIMBS as usize..2 * Fq::N_LIMBS as usize].to_vec());
+        let mut zero = Vec::new();
+        for _ in 0..8 {
+            zero.push(0u32);
+        }
+        let is_inf = (x == zero) & (y == zero);
```
eigmax: I would like to use x.is_zero() && y.is_zero() here.

PR #336, opened 2025-08-12
[Zellic Audit] blake3 fixes

### LINKED ISSUE #303: [Zellic Audit] blake3 final rounds (Avi 2 14.06)
blake3_utils::compress takes final_rounds as a parameter, and computes h'_i <- v'_i XOR v'_{i+8} for 0 <= i < final_rounds , which is correct for i < 8 , but not 8 <= i < 16 (which would have the formula h'_i <- v'_i XOR h_{i-8} . This seems not to be an issue currently since compress is only used with final_rounds = 8 .

Hakkush-07:
closes #303 and also implements opt suggestion (Avi 5)

### DIFF
--- a/bitvm/src/hash/blake3_utils.rs
+++ b/bitvm/src/hash/blake3_utils.rs
@@ -495,7 +495,7 @@ pub(crate) fn compress(
         let mut tmp = Vec::new();
 
         //iterate nibbles
-        for n in 0..8 {
+        for n in 0..final_rounds {
             let v2 = *state.get(&(i + 8)).unwrap();
             let v1 = state.get_mut(&i).unwrap();
             tmp.push(xor_2_nibbles(stack, v1, v2, 0, n, tables.use_full_tables));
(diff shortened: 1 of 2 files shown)


bitvm/src/hash/blake3_utils.rs
```diff
@@ -495,7 +495,7 @@ pub(crate) fn compress(
         let mut tmp = Vec::new();
 
         //iterate nibbles
-        for n in 0..8 {
+        for n in 0..final_rounds {
```
sander2, line 498: I'm not very familiar with this code, but wouldn't this push wrong values for `final_rounds > 8`? Since those values are supposed to be calculated differently? From the spec: 

image

eigmax: Agree. and the below expression `*state.get(&(i + 8)).unwrap();` would be panic due to accessing not-existed key.

I think we can add an assertion that final_rounds equals 8.

Hakkush-07:
closing for #366

PR #337, opened 2025-08-12
[Zellic Audit] Winternitz Fixes

just-erray:
This PR fixes the checksum length (it was subefficient previously) and, although unlikely it is to occur, adds a warning for repeated hashes, completing the work of #318. (Using a private key that has such feature would make it likely unsafe and unwanted, so fixing only the script doesn't fix the issue). 

There's also another potential bug caused by `rust-bitcoin-script` related to the usage of `bitcoin::Witness` for signatures, which will be clarified and fixed later. Due to that, this is currently a draft PR.

### DIFF
--- a/bitvm/src/signatures/winternitz.rs
+++ b/bitvm/src/signatures/winternitz.rs
@@ -150,7 +150,7 @@ pub trait Verifier {
             //        Maybe the script! macro removes the zeroes.
             //        There is a 1/256 chance that a signature contains a trailing zero.
             result.push(sig);
-            result.push(u32_to_le_bytes_minimal(digits[i as usize]));
+            result.push(bitcoin_representation(digits[i as usize]));
         }
         result
     }
--- a/bitvm/src/signatures/winternitz.rs
+++ b/bitvm/src/signatures/winternitz.rs
@@ -35,9 +35,9 @@ impl Parameters {
             message_digit_len,
             log2_base,
             checksum_digit_len: log_base_ceil(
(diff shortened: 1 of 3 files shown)


just-erray:
Paired with [this PR](https://github.com/BitVM/rust-bitcoin-script/pull/14), this should now fix the problem related to the usage of `bitcoin::Witness` for signatures.

bitvm/src/signatures/winternitz.rs
```diff
@@ -35,9 +35,9 @@ impl Parameters {
             message_digit_len,
             log2_base,
             checksum_digit_len: log_base_ceil(
-                ((1 << log2_base) - 1) * message_digit_len,
+                ((1 << log2_base) - 1) * message_digit_len + 1,
```
sander2, line 38: I think the +1 is not necessary. According to https://toc.cryptobook.us/book.pdf (chapter 14.3),  the checksum is  a numeric value with a max value of `(max_digit_value * message_digit_len`. The number of digits needed to represent that number is then `log_base_ceil(max_digit_value * message_digit_len, 1 << log2_base)`

For reference:

image

just-erray: I believe that you need the +1, since the representation of a number `x` in base `b` has `ceil(log_b(x + 1))` digits. For example, representation of 2 in binary has 2 digits instead of 1.

sander2: Ah yes my bad, you are right

PR #339, opened 2025-08-13
[Zellic Audit] fix fq6_hinted_mul_keep_elements

sevkett12:
I fixed **Mark 1**. Fq6::hinted_mul_keep_elements now works for all parameters (not only for (6,0)) to do that I removed hinted_mul_core and rewrote hinted_mul_keep_elements.

### DIFF
--- a/bitvm/src/bn254/fq6.rs
+++ b/bitvm/src/bn254/fq6.rs
@@ -218,37 +218,10 @@ impl Fq6 {
     // Input: [A, B]
     // Output: [C] where C = A x B
     pub fn hinted_mul(
-        a_depth: u32,
-        a: ark_bn254::Fq6,
-        b_depth: u32,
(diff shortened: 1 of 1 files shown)


bitvm/src/bn254/fq6.rs
```diff
@@ -422,6 +392,184 @@ impl Fq6 {
         (script, hints)
     }
 
+    // Input: [A, B]
+    // Output: [A, B, C] where C = A x B
+    pub fn hinted_mul_keep_elements(
```
eigmax, line 397: May I ask what's difference between hinted_mul and hinted_mul_keep_elements?

sevkett12: Hinted_mul consumes the stack elements whereas hinted_mul_keep_elements doesn't.

PR #342, opened 2025-08-17
fix #302 G1Affine::hinted_check_add has a soundness issue

### LINKED ISSUE #302: [Zellic Audit] G1Affine::hinted_check_add has a soundness issue (Mohit 2 11.06)

lynndell2010:
fix #302 G1Affine::hinted_check_add has a soundness issue


Hakkush-07:
#379 does this

PR #349, opened 2025-08-21
fix #348 The bigint multiplication can be optimized

### LINKED ISSUE #348: [Zellic Audit] 3.53 The bigint multiplication can be optimized
There are multiple redundant operations in BigIntImpl::mul (e.g., adding zero on the first iteration instead of conditionally emitting zero, using roll instead of a nonzero argument for double).

The byte size of U254::mul can be reduced from 116,218 to 111,466 (saving 4,752 bytes per multiplication), and the byte size of U64::mul can be reduced from 12,307 to 11,725 (saving 582 bytes per multiplication).

lynndell2010:
fix #348  The bigint multiplication can be optimized


Hakkush-07:
Since we dont use this function, I initially thought we could get rid of this, but we can instead merge this.

PR #366, opened 2025-08-21
fix #365 Blake3 script inscorrect

### LINKED ISSUE #365: [Zellic Audit] 3.27 Blake3 script inscorrect for more than 1,024 bytes of input or more than 32 bytes of output
The following excerpt from blake3_utils::init_state corresponds to assigning the words v12, v13, v14, v15 in the state to t_0, t_1, b, d in section 2.2 of the BLAKE3 specification.

state.push(stack.number_u32(0));
state.push(stack.number_u32(counter));
state.push(stack.number_u32(block_len));
state.push(stack.number_u32(flags));

In this implementation, counter is 32 bits; in the specification, t_0 is the low-order word of a 64-bit counter, and t_1 is the high-order word. Pushing 0 before counter is effectively treating counter as t_1 instead of as t_0, which is inconsistent with the specification's endianness. Additionally, the following excerpt from blake3_utils::compress corresponds to h′_i ←v′_i ⊕v′_{i+8}, which is only correct for 0 ≤i < 8. For 8 ≤i < 16, the specified formula is h′_i ←v′_i ⊕h_{i−8}, where h is the state prior to the application of the round functions.

Since this implementation is limited to one blake3 chunk (1,024 bytes), and t_1 · 2^64 + t_0 is a count in chunks, this does not currently (at the time of writing) affect correctness (i.e., counter is always 0 in the current implementation), but this will affect correctness if that limitation is ever lifted.

lynndell2010:
fix #365  Blake3 script inscorrect for more than 1,024 bytes of input or more than 32 bytes of output

### DIFF
--- a/bitvm/src/hash/blake3_utils.rs
+++ b/bitvm/src/hash/blake3_utils.rs
@@ -454,8 +454,8 @@ fn init_state(
     for u32 in &IV[0..4] {
         state.push(stack.number_u32(*u32));
     }
-    state.push(stack.number_u32(0));
     state.push(stack.number_u32(counter));
+    state.push(stack.number_u32(0));
     state.push(stack.number_u32(block_len));
     state.push(stack.number_u32(flags));
 


Hakkush-07:
Can you also add an assertion for final_round==8 to completely solve this finding?

Hakkush-07:
#390 does this

PR #379, opened 2025-09-11
[Zellic Audit] 3.12 Incomplete handling of G1 point addition

### LINKED ISSUE #302: [Zellic Audit] G1Affine::hinted_check_add has a soundness issue (Mohit 2 11.06)

Hakkush-07:
I created a different version of G1Affine::hinted_check_add.
- G1Affine::hinted_check_add_prevent_degenerate() is the old G1Affine::hinted_check_add, additionally some asserts that ensure the given points do not satisfy t==q or t==-q. The caller must use this function if they are sure of this.
- G1Affine::hinted_check_add() is a more general version. It handles the excluded cases above. However, this results in substantially larger script size since it includes both the addition and doubling parts together. (618763 >> 275192)

To achieve this, I added Fq::equal_keep_elements() function that simply checks equality without consuming the elements.

We continue to use the version that excludes these cases, because they do not occur currently.

Edit: Newly added function is optimized from 618763 to 481633 bytes. It now merges the same operations for doubling and adding. And a TODO is added to hints in this function so that later we must add hint validity checks.

closes #302

### DIFF
--- a/bitvm/src/bn254/g1.rs
+++ b/bitvm/src/bn254/g1.rs
@@ -264,16 +264,16 @@ impl G1Affine {
                         OP_FROMALTSTACK OP_DROP
                         { G1Affine::drop() }
                         { G1Affine::drop() }
-                        { G1Affine::push_zero() }
+                        { G1Affine::identity() }
                     OP_ELSE                                // case: t != -q
                         for _ in 0..Fq::N_LIMBS {
                             OP_DEPTH OP_1SUB OP_ROLL
                         }
-                        // TODO: uncomment out this line after 3.4 { Fq::copy(0) } { Fq::check_validity() }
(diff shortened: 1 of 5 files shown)


bitvm/src/bn254/g1.rs
```diff
@@ -212,6 +217,100 @@ impl G1Affine {
         (script, hints)
     }
 
+    pub fn hinted_check_add(t: ark_bn254::G1Affine, q: ark_bn254::G1Affine) -> (Script, Vec<Hint>) {
+        let mut hints = vec![];
+
+        let (alpha, bias) = if t == q {
```
eigmax: The first condition should exclude t == identity.

Hakkush-07: rearranged the if-checks with the same order as in the script

PR #395, opened 2025-09-23
Ensure FTPS uses TLS by enabling secure mode in Ftps driver

viktorking7:
### Description
- Problem: The FTPS driver initialized credentials with `is_secure=false`, resulting in plaintext FTP instead of FTPS.
- Change: Set `is_secure=true` in `bridge/src/client/data_store/ftp/ftps.rs` so FTPS connections use TLS.
- Impact: Credentials and data in transit are now encrypted; non‑TLS servers may fail to connect (expected for FTPS).

### DIFF
--- a/bridge/src/client/data_store/ftp/ftps.rs
+++ b/bridge/src/client/data_store/ftp/ftps.rs
@@ -37,7 +37,7 @@ impl Ftps {
         }
 
         let credentials = FtpCredentials {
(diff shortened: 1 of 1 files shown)


AaronFeickert:
Have you tested this change?

viktorking7:
> Have you tested this change?

yes, already tested

PR #405, opened 2025-10-18
Fix `is_positive` to correctly reference the last limb index

### LINKED ISSUE #403: `is_positive` picks the wrong limb
### Summary
`is_positive` picks the wrong limb: `{ (1 + depth) * Self::N_LIMBS } OP_PICK` should target the *last limb* at that depth, i.e., `… - 1`.

### Expected Behavior
Both helpers (`is_positive` and `is_negative`) should read the same head limb index (the last limb in the big integer at the given depth).

**Acknowledgments**: This issue was identified using Pomelo (https://eprint.iacr.org/2024/1768) by the UCSB/Nubit team.

trackoor:
Fix #403


trackoor:
not an issue

